PR #2, opened 2016-08-03
Initial setup

lilleyse:
I created a `Tools` directory that contains a Webstorm project for any of the command line tools we write such as gzipping, combining tilesets, etc. In future PRs there will be other folders for tileset generators and the validator. Right now each of these directories is like a self-contained npm module, but they all share the same git repo.


pjcozzi:
Add a root-level LICENSE.md even if it just says that all tools are licensed with Apache 2.0 and to check the LICENSE.md in each directory for each tool.

pjcozzi:
As for granularity, please only add three at this point:
* generators
* validator
* tools, e.g., like gltf-pipeline.  Do not put gzip/unzip, tileset.json fusion, etc. each into their own completely separate tool.  They should be seen as stages in a processing pipeline.

pjcozzi:
Also, what is the future layout?
```
Tools/validator
Tools/etc
```
or
```
validator
etc
```

pjcozzi:
We are also already mixing Pascal case and Camel case with the root `Tools` and `figures` directory.  I don't care which, but please keep them consistency except for validate exceptions.

lilleyse:
I was thinking this:
```
validator
etc
```

> tools, e.g., like gltf-pipeline. Do not put gzip/unzip, tileset.json fusion, etc. each into their own completely separate tool. They should be seen as stages in a processing pipeline.

Yes these will all be scripts in the bin folder of the Tools folder.

pjcozzi:
> I was thinking this:

OK, so this directory will be copied/renamed.  I see.

Let me know when this is ready to merge.

lilleyse:
Updated.

PR #3, opened 2016-08-03
Added gzipTileset script to tools

lilleyse:
* [x] Write tests
* [x] Promisify the module


pjcozzi:
Also:

* [ ] Update README.md with instructions.

lilleyse:
This is updated.

lilleyse:
Ok this is actually ready now.

tools/README.md
```diff
@@ -1 +1,20 @@
...
+Clone this repo and install [Node.js](http://nodejs.org/).  From the root directory of this repo, run:
+```
+npm install
+```
+
+## Tools
+
+### gzipTileset
+
+Detects whether the tileset is compressed or not and does the opposite.
```
pjcozzi: Are you sure this is the right form factor, it sounds awkward.

* I thought we were going to have one processing pipeline and be able to build a series of stages in exactly the same way as we plan to do so for gltf-pipeline.  Instead, this looks like the plan is to make lots of separate tools.  Are we sure that is the right approach?  If so, are we also taking the same approach with gltf-pipeline?  If so, what about the ability to provide a single JSON config to create the pipeline stages?
* Why not have separate zip / unzip arguments instead of "doing the opposite."  Did you look at other command-line zip tools?


> Detects whether the tileset

Is this precise?  I would expect it to test each tile so, for example, if 3 of 10 tests are already zipped, it only zips the other 7.

lilleyse: > I thought we were going to have one processing pipeline and be able to build a series of stages in exactly the same way as we plan to do so for gltf-pipeline.

Ah yeah, I wasn't really thinking about that when making this script but I'll start to support a pipeline approach. gltf-pipeline is about half-way there in terms of the grand vision.

> Why not have separate zip / unzip arguments instead of "doing the opposite." Did you look at other command-line zip tools?

I can do that.

> Is this precise? I would expect it to test each tile so, for example, if 3 of 10 tests are already zipped, it only zips the other 7.

I didn't really expect this to happen in practice, but it shouldn't be hard to support.

pjcozzi: The other option - that I would be OK with - is that each stage can be a separate command-line program and then the master pipeline command-line program would take the JSON config; however, this is different than, for example, `git`, `gzip`, etc.

pjcozzi:
@lilleyse is this ready now?

lilleyse:
Not yet, I still need to build out the command line tool/json input.

lilleyse:
Ok this is updated now. Let me know what you think about the pipeline JSON format.

pjcozzi:
> gunzip

ungzip?

pjcozzi:
> node ./bin/3d-tiles-tools.js gzip -i ./specs/data/TilesetOfTilesets/ -o ./output/TilesetOfTilesets-gzipped/

* Can `-i` be implicit or explicit?
* Can `-o` be optional?

Basically can I run this?

```
node ./bin/3d-tiles-tools.js gzip ./specs/data/TilesetOfTilesets/
```

pjcozzi:
Can you setup travis to run jsHint and the unit tests and coveralls for test coverage?  See [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline) for a simple example, but probably only 90% right with best practices.  Let me know if you need me to login to enable things.

pjcozzi:
Add instructions for running the tests and other npm scripts to the README.md.  You can somewhat copy from our other repos.

Basically, I want to get this repo fully setup for development while the code itself is so small and simple so we can easily onramp CIS 399 students in a few weeks.

pjcozzi:
The README.md is current written for advanced users.  Change the order of the presentation so the individual tools - and their options - are introduced first (e.g., so if I come to the repo asking "how do I gzip this tileset?" I can quickly find the easy way to do it), and then the pipeline is introduced for advanced users.

Also, in the next PR when the pipeline gets a second stage, please add a pipeline diagram so it is super clear how things work.

pjcozzi:
This is minor, but I wonder if there is any way to add a package.json to the root directory so that if someone runs `npm install` on the root directory like I did, we can display a message that says to run it on one of the sub-directories.  In general, you might want to look a bit at how repos with multiple node modules are designed.  I know it is common, but I don't know the best practices.

pjcozzi:
> Add instructions for running the tests and other npm scripts to the README.md. You can somewhat copy from our other repos.

For repoing coverage, I think you can exclude the `bin` directory.  See if we did so in `gltf-pipeline`.

pjcozzi:
There's a JSHint warning:
```
bin\3d-tiles-tools.js
  line 90  col 48  ['tilesOnly'] is better written in dot notation.

  ‼  1 warning
```

pjcozzi:
> Change the order of the presentation so the individual tools - and their options - are introduced first

Likewise, the pipeline JSON section should show all of each stages options perhaps using formatting similar to the glTF spec.  The glTF spec is actually generated from the JSON schema, but at this point, this does not need one.

pjcozzi:
> node ./bin/3d-tiles-tools.js pipeline -i ./specs/data/pipeline.json

Can we make this as simple as
```
node ./bin/3d-tiles-tools.js ./specs/data/pipeline.json
```

tools/README.md
```diff
@@ -1 +1,68 @@
...
+Executes a pipeline JSON file containing an input path, output path, and list of stages to run.
+A stage can be a string specifying the stage name or an object specifying the stage name and any additional parameters.
+Stages are executed in the order listed.
+
+This example `pipeline.json` gzips the input tilest and saves it in the given output directory.
+
+```json
+{
+    "input": "Tileset/",
+    "output": "TilesetGzipped/",
```
pjcozzi: Is `output` optional?  If not, can it be?

tools/index.js
```diff
@@ -1,2 +1,4 @@
 module.exports = {
+    gzipTileset : require('./lib/gzipTileset'),
```
pjcozzi: Perhaps just submit an issue for now to generate doc for these.

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+var defaultValue = Cesium.defaultValue;
+var defined = Cesium.defined;
+var DeveloperError = Cesium.DeveloperError;
+
+module.exports = gzipTileset;
+
+/**
+ * gzips or gunzips the input tileset.
+ *
+ * @param {String} inputDirectory Path to the input directory.
```
pjcozzi: Consider putting this all in the `options` object (like Cesium) or maybe strictly follow a pattern where required parameters are not in `options` and required ones are.  Ask I mentioned above, I think `outputDirectory` should be optional.

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {Object} [outputDirectory] Path to the output directory.
+ * @param {Object} [options] Object with the following properties:
+ * @param {Boolean} [options.gzip=true] Whether to gzip or gunzip the tileset.
```
pjcozzi: Should this have `options.overwrite=false` (exposed throughout, of course) so users don't accidentally overwrite with their output?

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+function isTile(file) {
+    var extension = path.extname(file);
+    return extension === '.b3dm' ||
+           extension === '.i3dm' ||
+           extension === '.pnts' ||
+           extension === '.cmpt' ||
+           extension === '.vctr';
+}
+
+function isGzipped(file) {
```
pjcozzi: This could be in its own `@private` file instead of duplicating it in the tests.

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+                            return fsExtraEnsureDir(path.dirname(outputFile))
+                                .then(function () {
+                                    var inp = fsExtra.createReadStream(inputFile);
+                                    var out = fsExtra.createWriteStream(outputFile);
+                                    var operation = gzip ? zlib.createGzip() : zlib.createGunzip();
+                                    return streamToPromise(inp.pipe(operation).pipe(out));
+                                });
+
+                        });
+                }, {concurrency: 1024})
```
pjcozzi: Are you sure about 1024?

Should this be an option?  I don't want to over optionify things, but if there isn't a one size fits all when this is integrated into a more complex pipeline and server, this should probably be configurable.

tools/lib/runPipeline.js
```diff
@@ -0,0 +1,112 @@
...
+        path.join(path.dirname(inputPath), path.basename(inputPath) + '-processed')));
+
+    if (!defined(stages)) {
+        return fsExtraCopy(inputPath, outputPath);
+    }
+
+    options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+    var verbose = defaultValue(options.verbose, false);
+
+    var workingDirectory1 = path.join(path.dirname(inputPath), path.basename(inputPath) + '-working1');
```
pjcozzi: Here and below would be more robust if we use a working directory name that we know doesn't exist.

pjcozzi:
If this were ran on a server, are there any security concerns we need to address now, e.g., reading random directories, etc.

pjcozzi:
Just those comments, this project is off to a nice start!

tools/lib/gzipTileset.js
lilleyse: In my experience it's always a hassle to remember to set overwrite to true if the default is false. Maybe we can give the option but set default to true? Is that even worth doing?

lilleyse:
> Can you setup travis to run jsHint and the unit tests and coveralls for test coverage? See gltf-pipeline for a simple example, but probably only 90% right with best practices. Let me know if you need me to login to enable things.

You'll need to do this, I don't have admin access.

> This is minor, but I wonder if there is any way to add a package.json to the root directory so that if someone runs npm install on the root directory like I did, we can display a message that says to run it on one of the sub-directories. In general, you might want to look a bit at how repos with multiple node modules are designed. I know it is common, but I don't know the best practices.

I'm not sure if it's possible to override `npm install`.

> If this were ran on a server, are there any security concerns we need to address now, e.g., reading random directories, etc.

Yeah we would need to handle the input and output directories given in pipeline.json. I'll look into this some more.

> Output directory optional

Just replying to a lot of related comments here, the output directory is always optional.

lilleyse:
Made some changes, still have a few more things to round out.

tools/lib/gzipTileset.js
pjcozzi: Just follow whatever the Unix philosophy is.

lilleyse: Many unix command line tools do ask for permission before overwriting. Node libraries on the other hand like `fs` always overwrite. If I were to chose I prefer the overwrite always way.

tools/lib/gzipTileset.js
lilleyse: I didn't give much thought to it, I suppose keeping the number like 1024 will prevent too many promises from getting created, but still keep the operation fast. I'm not really sure what the best values are for a server or if it matters. @mramato thoughts?


and am not really sure of the implications of it for a server.

lilleyse:
> If this were ran on a server, are there any security concerns we need to address now, e.g., reading random directories, etc.

I believe it should be safe as long as the server that wraps this project writes the input files and output files to temp directories before processing, as well as changes the relative paths in the pipeline JSON to absolute paths (like in `3d-tiles-tools.js`). I'm not sure if there's much we can do here.

tools/lib/gzipTileset.js
pjcozzi: What about Node cli apps?  It's fine for the library to do one thing and the cli to do the other, I suppose, let's just make sure we are doing what is standard.

pjcozzi:
>> Can you setup travis to run jsHint and the unit tests and coveralls for test coverage? See gltf-pipeline for a simple example, but probably only 90% right with best practices. Let me know if you need me to login to enable things.
> You'll need to do this, I don't have admin access.

I'll submit an issue; it will be a bit before I have time for this.

Anything else left for this PR?

lilleyse:
> Anything else left for this PR?

This should be ready now. `-f, --force` is now added to overwrite existing folders.

lilleyse:
The only unresolved comment is https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/3#discussion_r75106486.

tools/lib/gzipTileset.js
mramato: For a cli program, 1024 is fine, but if you want to be conservative, a number like 100 should work fine too and I doubt you would see a performance difference since you are likely to starve the drive you're reading from in both cases.

This function would need a lot of work before I would consider using it in a server environment.  At a quick glance:

1. You wouldn't buffer the entire list of files in memory up front, that would cause problems for huge tilesets
2. Output would never go to disk, instead I would expect a callback to be raised for each tile as it is zipped and let the caller decide where the data gets written.
3. You wouldn't call `console.log` directly, logging/progress would be the of the callback in 1.

tools/bin/3d-tiles-tools.js
```diff
@@ -0,0 +1,96 @@
...
+var runPipeline = require('../lib/runPipeline');
+
+var fsExtraReadJson = Promise.promisify(fsExtra.readJson);
+
+var defaultValue = Cesium.defaultValue;
+var defined = Cesium.defined;
+
+if (process.argv.length < 4 || defined(argv.h) || defined(argv.help) || !defined(argv._[0])) {
+    var help =
+        'Usage: node ' + path.basename(__filename) + ' command [args]\n' +
```
mramato: `node` should not be here.  this repository should be installed with `-g` at which point every command would be in the user's system path.

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+    return new Promise(function(resolve, reject) {
+        options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+        var gzip = defaultValue(options.gzip, true);
+        var tilesOnly = defaultValue(options.tilesOnly, false);
+        var verbose = defaultValue(options.verbose, false);
+
+        if (!defined(inputDirectory)) {
+            reject(new DeveloperError('inputDirectory is required'));
+        }
+        inputDirectory = path.normalize(inputDirectory);
```
mramato: `path.normalize` for this and the next line really belong in the calling code, not here.

tools/lib/gzipTileset.js
```diff
@@ -0,0 +1,110 @@
...
+ */
+function gzipTileset(inputDirectory, outputDirectory, options) {
+    return new Promise(function(resolve, reject) {
+        options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+        var gzip = defaultValue(options.gzip, true);
+        var tilesOnly = defaultValue(options.tilesOnly, false);
+        var verbose = defaultValue(options.verbose, false);
+
+        if (!defined(inputDirectory)) {
+            reject(new DeveloperError('inputDirectory is required'));
```
mramato: Because it's a `DeveloperError`, this check should happen outside of the Promise and throw immediately.

mramato:
> The only unresolved comment is #3 (comment).

Just to clarify, I don't think you necessarily needy to make any of those changes.  I was just trying to illustrate how writing server-side tools is very different than writing cli tools.  Of course if our goal is to provide a library of tools that will be used either on the client or the server, then we need to lay out some architecture ground rules up front.

pjcozzi:
>> The only unresolved comment is #3 (comment).
> Just to clarify, I don't think you necessarily needy to make any of those changes. I was just trying to illustrate how writing server-side tools is very different than writing cli tools. Of course if our goal is to provide a library of tools that will be used either on the client or the server, then we need to lay out some architecture ground rules up front.

If this is easy enough to address now, I would like to do it soon.

@mramato please keep a backlog of updates for the Node Guide.

lilleyse:
I addressed https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/3#r77263777.

Now the tiles are gzipped during the directory walk rather than all at once. To check that all the files are processed, I keep track of the current count and compare that with the total number of files in the directory.

I also added writing and logging callbacks. However I'm not completely sure what to do with the write callbacks when there are multiple stages. Currently intermediary stages ping-pong between two temp directories. Should the write callback have control over those too? I have a TODO in the code for this.

@mramato I would appreciate your thoughts on the current approaches.

pjcozzi:
@mramato if you have any more feedback, please submit an issue.

Issue #1, opened 2016-08-03
3D Tiles Tools roadmap


pjcozzi:
More ideas to come...

# Tools

**Definitely**

* [ ] Fusion: Merge tileset with external tileset.json references into one tileset.json
* [ ] Fission: Separate a monolithic tileset.json into several tileset.json based on different criteria: e.g., `n` tiles per tileset.json, `m` levels per tileset.json, etc.
* [ ] Generate stats like [gltf-statistics](https://github.com/AnalyticalGraphicsInc/gltf-statistics).

**As Needed**

* [ ] Visualize a tileset's spatial data structure, perhaps even an interactive tileset explorer using Cesium with D3 overlay
* [ ] Separate composites, i.e., separate a composite tile into several tiles
   * [ ] Separate out into separate tilesets if useful, e.g., one for buildings, one for trees, etc.
   * [ ] Combine tiles into composites
* [ ] Merge multiple tilesets into one tileset

# Validator

* [ ] Detailed list TBA
* [ ] Integrate [glTF Validator](https://github.com/KhronosGroup/glTF-Validator) for `b3dm` and `i3dm` tile formats

PR #4, opened 2016-08-19
Make composite tile

lasalvavida:
Adds a tool for generating composite tilesets.


generator/bin/makeCompositeTile.js
```diff
@@ -0,0 +1,56 @@
...
+'use strict';
+
+var Cesium = require('cesium');
+var Promise = require('bluebird');
+var fs = require('fs-extra');
+
+var defined = Cesium.defined;
+
+var argv = require('yargs')
+    .usage('Usage: $0 \<tilesets\> [options]')
```
pjcozzi: "tilesets" -> "tiles", right?

pjcozzi:
Thanks @lasalvavida!  Please add doc to the README.md, and add unit tests.

@lilleyse please review and merge when this is ready.

generator/bin/makeCompositeTile.js
lasalvavida: Yup! Thank you

pjcozzi:
@lasalvavida @lilleyse also please keep the roadmap, #1, updated when things are merged.  I'll leave this 100% up to you guys.

lasalvavida:
Updated

generator/README.md
```diff
@@ -1 +1,17 @@
-Tools for generating sample 3D Tiles tilesets.
\ No newline at end of file
+Tools for generating sample 3D Tiles tilesets.
+
+# Make Composite Tile
+
+Creates a composite tile from multiple source tiles.
+
+## Example
+`node ./bin/makeCompositeTile.js batched.b3dm instanced.i3dm points.pnts -o output/composite.cmpt`
```
pjcozzi: Does this work with input wildcards?  Not a huge deal, but would be fine if it is trivial to implement.

pjcozzi: Is the `output` directory created if it doesn't exist?  I am not familiar with all the Node IO routines.

generator/bin/makeCompositeTile.js
```diff
@@ -0,0 +1,56 @@
...
+    .demand(1)
+    .argv;
+
+var readTile = require('../lib/readTile');
+var writeTile = require('../lib/writeTile');
+
+var tilePaths = argv._;
+var outputPath = argv.o;
+
+Promise.map(tilePaths, function(tilePath) {
```
pjcozzi: Can we move this logic to a function in `lib` so it could be part of an API if we decide to expose one later?

pjcozzi:
Looks great at a quick glance!  @lilleyse review and merge please.

generator/README.md
lasalvavida: > Does this work with input wildcards? Not a huge deal, but would be fine if it is trivial to implement.

It does not, I'll see how hard that is to do.

lasalvavida: > Is the output directory created if it doesn't exist? I am not familiar with all the Node IO routines.

It does because `writeTile` uses `fs-extra`.

generator/bin/makeCompositeTile.js
lasalvavida: Which part? The `Promise.map` part is pretty trivial.

generator/README.md
lasalvavida: >> Does this work with input wildcards? Not a huge deal, but would be fine if it is trivial to implement.

>It does not, I'll see how hard that is to do.

Actually I just tried it and it does work. Your shell handles the wildcards and passes the resolved arguments.

generator/bin/makeCompositeTile.js
pjcozzi: I think all of it so it is a function that takes an array of input paths and an output path.

lasalvavida: How about just separating out the composite logic? Make it a function that takes an array of buffers and outputs the composite tile buffer. I'd rather not force it to be to/from the filesystem.

pjcozzi: OK

lasalvavida:
Updated

pjcozzi:
Thanks @lasalvavida.  @lilleyse can you please do the final review and merge?

lilleyse:
Sure, I haven't reviewed this at all yet but I'm going to take some time later to review this and finish up #3

generator/specs/bin/makeCompositeTileSpec.js
```diff
@@ -0,0 +1,83 @@
...
+    var makeCompositeTileProcess = childProcess.fork('./bin/makeCompositeTile', args);
+    return new Promise(function(resolve, reject) {
+        makeCompositeTileProcess.on('close', function() {
+            resolve();
+        });
+        makeCompositeTileProcess.on('error', function() {
+            reject();
+        });
+    });
+}
```
lilleyse: I left a similar [comment](https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/6#discussion_r81565846) on #6, we aren't running tests on the bin folder right now. However if you want to you write tests for `3d-tiles-tools.js` and include this code there feel free to do so.

generator/package.json
```diff
@@ -37,7 +38,7 @@
     "open": "0.0.5",
     "request": "2.74.0",
     "requirejs": "2.2.0",
-    "yargs": "4.8.1"
+    "rimraf": "2.5.4"
```
lilleyse: You can use `fsExtra.remove` instead.

generator/lib/readTile.js
```diff
@@ -0,0 +1,38 @@
...
+    return Promise.resolve()
+        .then(function() {
+            if (!defined(filePath)) {
+                throw new DeveloperError('filePath must be defined');
+            }
+            return fsReadFile(filePath);
+        })
+        .then(function(buffer) {
+            if (isGzipped(buffer)) {
+                return zlibGunzip(buffer, undefined);
```
lilleyse: No need to pass in undefined.

generator/lib/writeTile.js
```diff
@@ -0,0 +1,47 @@
...
+function writeTile(filePath, tileData, options) {
+    return Promise.resolve()
+        .then(function() {
+            if (!defined(filePath)) {
+                throw new DeveloperError('filePath must be defined.');
+            }
+            if (!defined(tileData)) {
+                throw new DeveloperError('tileData must be defined.');
+            }
+            options = defaultValue(options, {});
```
lilleyse: `options = defaultValue(options, defaultValue.EMPTY_OBJECT);`

generator/lib/writeTile.js
```diff
@@ -0,0 +1,47 @@
...
+            var gzip = defaultValue(options.gzip, false);
+            if (gzip) {
+                return zlibGzip(tileData, undefined);
```
lilleyse: I think there was some talk that `gzip` is faster as a sync operation because it's not actually doing any file reading. @mramato is this correct?

generator/specs/lib/isGzippedSpec.js
```diff
@@ -0,0 +1,23 @@
...
+var zlib = require('zlib');
+var isGzipped = require('../../lib/isGzipped');
+
+var zlibGzip = Promise.promisify(zlib.gzip);
+
+describe('isGzipped', function() {
+    it('throws DeveloperError if data is undefined', function() {
+        expect(function() {
+            isGzipped(undefined);
+        }).toThrowError();
```
lilleyse: Change to `toThrowDeveloperError`

generator/specs/lib/isGzippedSpec.js
```diff
@@ -0,0 +1,23 @@
...
+    });
+
+    it('detects when data is gzipped', function(done) {
+        var data = new Buffer(40);
+        expect(isGzipped(data)).toBeFalsy();
```
lilleyse: `toBe(false)` and `toBe(true)` are more precise.

generator/specs/lib/writeTileSpec.js
```diff
@@ -0,0 +1,57 @@
...
+        var path = testOutputPath + 'justmagic.i3dm';
+        var data = new Buffer('i3dm');
+        expect(writeTile(path, data)
+            .then(function() {
+                return fsReadFile(path);
+            })
+            .then(function(tileData) {
+                var magic = tileData.toString('utf8', 0, 4);
+                expect(magic).toEqual('i3dm');
+            }),done).toResolve();
```
lilleyse: `, done` spacing

generator/specs/lib/writeTileSpec.js
```diff
@@ -0,0 +1,57 @@
...
+    it('writes a gzipped tile', function(done) {
+        var path = testOutputPath + 'justmagic.i3dm';
+        var data = new Buffer('i3dm');
+        expect(writeTile(path, data, {gzip : true})
+            .then(function() {
+                return fsReadFile(path);
+            })
+            .then(function(tileData) {
+                expect(isGzipped(tileData)).toBeTruthy();
+            }),done).toResolve();
```
lilleyse: Spacing

lilleyse:
Add `makeCompositeTile` to `index.js`.

lasalvavida:
@lilleyse This is ready for a look.

tools/lib/getDefaultWriteCallback.js
```diff
@@ -1,18 +1,13 @@
...
 /**
  * @private
  */
-function getDefaultWriteCallback(outputDirectory) {
-    return function(file, data) {
-        var outputFile = path.join(outputDirectory, file);
-        return fsExtraOutputFile(outputFile, data);
+function getDefaultWriteCallback() {
+    return function(file, data, options) {
+        return writeTile(file, data, options);
```
lilleyse: I think it's best to just keep this really simple like it was before where it just gets data and write to a file. Gzipping or anything else should be handled by the calling code. Maybe this makes `writeTile` not needed anymore.

tools/lib/gzipTileset.js
```diff
@@ -40,100 +41,26 @@ function gzipTileset(options) {
...
-}
-
-function getNumberOfFilesInDirectory(directory) {
-    return new Promise(function(resolve, reject) {
-        var numberOfFiles = 0;
-        fsExtra.walk(directory)
-            .on('data', function (item) {
-                if (!item.stats.isDirectory()) {
-                    ++numberOfFiles;
+    return getFilesInDirectory(inputDirectory, {
```
lilleyse: I used to do a similar approach here, but changed it for this comment: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/3#r77263777. I'm fine with merging the changes in this file in though.

I have a branch that cleans up `gzipTileset` and the end result looks similar to this, but walks the directory without buffering the file names up front (through a helper function). And really that too is probably going to get blown away again at some later point.

tools/lib/isTileFile.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+var path = require('path');
+
+module.exports = isTileFile;
+
+function isTileFile(file) {
```
lilleyse: Mark as @private

tools/specs/lib/gzipTilesetSpec.js
```diff
@@ -173,21 +172,20 @@ describe('gzipTileset', function() {
...
     it('accepts custom writeCallback that returns a promise', function (done) {
-        var outputDirectory = gzippedDirectory;
         var writeCallback = function(file, data) {
-            var outputFile = path.join(outputDirectory, file);
-            return fsExtraOutputFile(outputFile, data);
+            return fsExtraOutputFile(file, data);
         };
         var gzipOptions = {
             inputDirectory : tilesetDirectory,
+            outputDirectory : gzippedDirectory,
```
lilleyse: The idea behind custom write callbacks was that you didn't need to supply an output directory to the stage because that information only needs to be known by the callback. I think the old way with using a closure for `getDefaultWriteCallback` was more flexible.

lilleyse:
Also add `makeCompositeTile` to the README and `3d-tiles-tools.js` Just make sure it's clearly defined that its not a pipeline stage and doesn't operate over the whole tileset.

lilleyse:
Brought the `makeCompositeTile` file into https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/4

PR #11, opened 2016-10-04
Validator - delete writeTile

lilleyse:
Removed writeTile since we won't need this in the validator.


pjcozzi:
Let's run this project like any other repo and try to avoid insta-merging unless it is really blocking.

Issue #10, opened 2016-10-04
Validator - readTileset


lilleyse:
@sumitshyamsukha

Write a lib file called `readTileset` that loads a `tileset.json` file, check that the contents are valid JSON, and return the JSON object. Throw an error if the JSON is not valid.

sumitshyamsukha:
Don't have access to push yet -- could I be granted access? @lilleyse

lilleyse:
Thanks for the reminder, you and @JudyWeng are added now.

FreakTheMighty:
Sorry to bring back a very old ticket, but I wonder if it would be worth using something a json-schema library like [this](https://ajv.js.org/) to handle basic validation.

PR #13, opened 2016-10-05
Added readTileset

sumitshyamsukha:
#10 
@lilleyse Let me know if I should change anything!


validator/lib/readTileset.js
```diff
@@ -0,0 +1,15 @@
...
+var fsReadFile = Promise.promisify(fs.readJson);
+
+function readTileset(filePath) {
+    return fsReadFile(filePath)
+        .then(function (jsonObject) {
+            return jsonObject;
+        })
+        .catch(function(error)  {
+            throw error;
+        });
```
lilleyse: No need to catch and then throw the error. The error will propagate with the returned promise. Anything that calls `readTileset` would then be responsible for catching the error:

```javascript
readTileset('tileset.json')
    .then(function(json) {

    })
    .catch(function(error) {
        console.log('there was an error')
    });
}
```

validator/lib/readTileset.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+var fs = require('fs-extra');
+var Promise = require('bluebird');
+
+var fsReadFile = Promise.promisify(fs.readJson);
+
+function readTileset(filePath) {
```
lilleyse: You will have to rework this function to support checking if the file is gzipped or not. Check out `readTile` for an example.

validator/lib/readTileset.js
```diff
@@ -0,0 +1,15 @@
...
+    return fsReadFile(filePath)
```
lilleyse: `fsReadFile` returns a Buffer, but you are assuming it's a JSON object. It's good to have the buffer though because then you can check if it's gzipped. If it is gzipped you will need to decode the buffer, convert to a string, and then call `JSON.parse`. If it isn't gzipped, you can skip the decode step.

validator/lib/readTileset.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
lilleyse: I didn't put this in the issue, but can you write tests for this function? Look at `readTileSpec` for an example. A couple things to check:
* a normal JSON file
    * Use `specs/data/Tileset/tileset.json` as your test case
* a gzipped JSON file
    * We already have a tileset at `specs/data/Tileset`. Create a gzipped tileset at `specs/data/TilesetGzipped`. You can use the gzip utility in https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/tree/master/tools to generate a gzipped tileset from `specs/data/Tileset`. Then use `specs/data/TilesetGzipped/tileset.json` as your test case.
* a file that isn't valid JSON
    * look at https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/tools/specs/lib/gzipTilesetSpec.js#L155 for an example of checking a rejected promise
    * As a test file you can use one of the b3dm's in the `specs/data/Tileset` folder

validator/lib/readTileset.js
lilleyse: Oh sorry, I didn't realize you promisified the readJson function based on the naming. Still it's a good idea to follow these steps.

sumitshyamsukha:
@lilleyse I'll look into making the tests, but for now this PR has been updated to reflect gzipping.

validator/lib/readTileset.js
```diff
@@ -0,0 +1,24 @@
...
+var fs = require('fs-extra');
+var Promise = require('bluebird');
+var zlib = require('zlib');
+var isGzipped = require('./isGzipped');
+
+var fsReadFile = Promise.promisify(fs.readFile);
+var zlibGunzip = Promise.promisify(zlib.gunzip);
+var parseJson = Promise.promisify(JSON.parse)
```
lilleyse: No need to promisify `JSON.parse` because it is a synchronous function.

lilleyse:
The updates looks good. I'll do a final review once the tests are in.

validator/lib/readTileset.js
sumitshyamsukha: @lilleyse Just a quick question for the first two test cases: what would I put within the function once I've received the parsed JSON such as in `readTileSpec`?

lilleyse: Maybe just check that json.asset is defined since every tileset has an "asset" section. This should be good enough.

sumitshyamsukha:
@lilleyse Added the test. Please let me know if there's anything else to fix!

validator/lib/readTileset.js
sumitshyamsukha: `Running gzip
Compressing files...
Unhandled rejection Error: EISDIR: illegal operation on a directory, open 'output/TilesetOfTilesets-gzipped'
    at Error (native)`

This is what I receive on trying to gzip the tileset. Any advice on how to fix this?

lilleyse: What is the exact command you are doing?

sumitshyamsukha: `node ./bin/3d-tiles-tools.js gzip ./specs/data/TilesetOfTilesets/tileset.json ./output/TilesetOfTilesets-gzipped/`

sumitshyamsukha:
@lilleyse Updated the tests, please let me know if it's all good.

validator/specs/lib/readTilesetSpec.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+var readTileset = require('../../lib/readTileset');
+
+describe('readTileset', function() {
+    it('reads a tileset', function(done) {
+        expect(readTileset('./specs/data/Tileset/tileset.json')
+            .then(function(tilesetData) {
+                return (tilesetData['asset'] !== undefined)
```
lilleyse: Use `defined` instead for this and below. Example: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/validator/specs/matchers/expectPromise.js

lilleyse:
Just that comment. The tests looks good.

validator/specs/lib/readTilesetSpec.js
sumitshyamsukha: @lilleyse Added `defined` instead. Please let me know if it's okay.

validator/specs/lib/readTilesetSpec.js
```diff
@@ -1,18 +1,29 @@
...
 describe('readTileset', function() {
     it('reads a tileset', function(done) {
         expect(readTileset('./specs/data/Tileset/tileset.json')
             .then(function(tilesetData) {
-                return (tilesetData['asset'] !== undefined)
+                if (defined(tilesetData['asset'])) {
+                    done();
+                } else {
+                    throwError("not a tileset");
+                }
```
sumitshyamsukha: Updated, should be good to go now?

validator/lib/readTileset.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+var fs = require('fs-extra');
+var Promise = require('bluebird');
+var zlib = require('zlib');
+var isGzipped = require('./isGzipped');
+
+var fsReadFile = Promise.promisify(fs.readFile);
+var zlibGunzip = Promise.promisify(zlib.gunzip);
+
+/**
```
lilleyse: This file needs to include `module.exports = readTileset;`

validator/specs/lib/readTilesetSpec.js
```diff
@@ -0,0 +1,23 @@
...
+describe('readTileset', function() {
+    it('reads a tileset', function(done) {
+        expect(readTileset('./specs/data/Tileset/tileset.json')
+            .then(function(tilesetData) {
+                expect(tilesetData).toBeDefined();
+            }), done).toResolve();
+    });
+
+    it('reads a gzipped tileset', function(done) {
+        expect(readTile('./specs/data/TilesetGzipped/tileset.json')
```
lilleyse: `readTile` -> `readTileset`

validator/lib/readTileset.js
```diff
@@ -0,0 +1,29 @@
...
+ * Reads tileset data from a file.
+ *
+ * @param {String} filePath The file path to read from.
+ * @returns {JSON} A JSON that contains the parsed data in the filePath.
```
lilleyse: This function return a promise, not JSON. Check out the description in `readTile`.

lilleyse:
I believe these are the last couple of issues. Make sure to run `jsHint` and `test` regularly to catch some of these as you go.

validator/lib/readTileset.js
sumitshyamsukha: I thought JSON.parse would return a JSON?

lilleyse: Yeah it does, but technically `readTileset` returns a promise that contains JSON and so it should be worded that way.

sumitshyamsukha:
Updated, let me know if there's anything else!

lilleyse:
Ok! This all looks good now. I made a small change to `readTileSpec`, but that's unrelated to the code here. Thanks @sumitshyamsukha.

Issue #9, opened 2016-10-04
Validator - validateB3dm header


lilleyse:
@JudyWeng

Write a lib file called `validateB3dm` that takes a Buffer containing the contents of a b3dm (like as returned by `readTile`) and checks that it is valid according the the [b3dm spec](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel). For now just check that:

magic is "b3dm"
version is 1
byteLength is equal to the Buffer's length
batchTableJSONByteLength - skip for now, will require more validation later
batchTableBinaryByteLength - skip for now, will require more validation later
batchLength - skip for now, this is dependent on the glTF within the b3dm

This issue will get more involved and will eventually extend to the other tile formats.

JudyWeng:
I tried to push earlier and it says I do not have access to 3d-tiles-tools. Am I supposed to push elsewhere?

lilleyse:
Did you receive any emails about being added to the organization?

JudyWeng:
oh sorry, yeah, it went to my spam box for some reason. Thanks!

lilleyse:
Fixed in #14

lilleyse:
And #16

PR #14, opened 2016-10-08
Validate b3dm

JudyWeng:
I'm not sure if I did this correctly. Please let me know if there is anything I need to fix.
#9


validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
```
lilleyse: Fine to remove this whitespace.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+var Cesium = require('cesium');
+var DeveloperError = Cesium.DeveloperError;
+module.exports = validateB3dm;
```
lilleyse: Add empty lines between these three, this will provide better styling once we start to include more libraries. Look at `readTile` for example.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Test if the provided data is gzipped.
```
lilleyse: Change this description.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @param {Buffer} content A buffer containing the contents of a b3dm tileset.
```
lilleyse: tileset -> tile

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+ * @returns {Boolean} True if the content is valid according to the specs:
```
lilleyse: Change specs to spec.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+ * https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel, False if not.
```
lilleyse: Use @link.

For example: `{@link http://www.khronos.org/registry/webgl/specs/latest/#5.2|WebGLContextAttributes}`

Also run `npm run jsDoc` to verify that the documentation looks good.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+function validateB3dm(content) {
+    if(content === undefined || content === null) {
```
lilleyse: Use `Cesium.defined`, which does both of these checks.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+        throw new DeveloperError('b3dm content is undefined or null');
```
lilleyse: Change the description so its a bit more directive, like `content must be defined`.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    //buf.writeUInt32LE(value, offset[, noAssert])
+    //buf.readUInt32LE(offset[, noAssert])
```
lilleyse: Remove these lines.

validator/lib/validateB3dm.js
```diff
@@ -0,0 +1,26 @@
...
+
+    var magic = content.toString('uft8', 0, 4);
+    var version = content.readUInt32LE(4);
+    var byteLength = content.readUInt32LE(8);
+
+    return (magic === 'bd3m' && version === 0x01 && byteLength === content.length);
```
lilleyse: Change `0x01` to `1`.

This approach is fine for now, but once we do more checking in this function each of these statements could be its own if statement that returns false if the condition isn't met.

JudyWeng:
Sorry for waiting so long to get back about this issue. I tried to start the validateB3dmSpec file and I've been reading about Promises and readTileSpec, but I am still a bit confused on what readTileSpec and validateB3dmSpec is supposed to do..?

I also just made a js file with the methods that would return invalid b3dm tiles, but I am not sure how I would use this to actually generate the tiles and test the method against the these tiles.

lilleyse:
Closed in favor of #16

PR #18, opened 2016-10-26
Initial validator

sumitshyamsukha:
@lilleyse Still not complete, had a couple of questions, but this is the basic structure

### DIFF
--- a/validator/bin/3d-tiles-validator.js
+++ b/validator/bin/3d-tiles-validator.js
@@ -9,8 +9,8 @@ var validateTileset = require('../lib/validateTileset');
 var filePath = argv._[0];
 readTileset(filePath)
     .then(function(json) {
-       validateTileset(json)
-           .then(function(response) {
-            console.log(response.message);
-        })
+        validateTileset(json)
+            .then(function(response) {
+                console.log(response.message);
+            });
     });
--- a/validator/bin/3d-tiles-validator.js
+++ b/validator/bin/3d-tiles-validator.js
@@ -1,2 +1,15 @@
 #!/usr/bin/env node
 'use strict';
+
+var argv = require('yargs').argv;
+var fsExtra = require('fs-extra');
+var path = require('path');
+var Promise = require('bluebird');
+
+var readTileset = require('../lib/readTileset');
+var validateTileset = require('../lib/validateTileset');
+
+var filePath = argv._[0];
+var jsonTileset = readTileset(filePath);
+var response = validateTileset(jsonTileset);
+console.log(response);
--- a/validator/bin/3d-tiles-validator.js
+++ b/validator/bin/3d-tiles-validator.js
@@ -2,14 +2,15 @@
 'use strict';
 
 var argv = require('yargs').argv;
-var fsExtra = require('fs-extra');
-var path = require('path');
-var Promise = require('bluebird');
 
 var readTileset = require('../lib/readTileset');
 var validateTileset = require('../lib/validateTileset');
 
 var filePath = argv._[0];
-var response = validateTileset(readTileset(filePath));
-
-console.log(response.message);
+readTileset(filePath)
+    .then(function(json) {
+       validateTileset(json)
+           .then(function(response) {
+            console.log(response.message);
+        })
+    });
--- a/validator/bin/3d-tiles-validator.js
+++ b/validator/bin/3d-tiles-validator.js
@@ -10,6 +10,6 @@ var readTileset = require('../lib/readTileset');
 var validateTileset = require('../lib/validateTileset');
 
 var filePath = argv._[0];
-var jsonTileset = readTileset(filePath);
-var response = validateTileset(jsonTileset);
-console.log(response);
+var response = validateTileset(readTileset(filePath));
+
+console.log(response.message);
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var Promise = require('bluebird');
-var Cesium = require('Cesium');
+var Cesium = require('cesium');
 var defined = Cesium.defined;
 
 module.exports = validateTileset;
@@ -25,10 +25,9 @@ function validateNode(root, parent, resolve) {
     stack.push({
           node: root,
           parent: parent
-        });
+    });
 
     while (stack.length > 0) {
-
         var node = stack.pop();
         var tile = node.node;
         var parent = node.parent;
@@ -49,9 +48,7 @@ function validateNode(root, parent, resolve) {
                 });
             }
         }
-
     }
-
     return resolve({
         result : true,
         message : 'Tileset is valid'
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -23,10 +23,11 @@ function validateNode(node, parentGeometricError, resolve) {
     if (node.geometricError > parentGeometricError) {
         return resolve(false, 'Child has geometricError greater than parent');
     }
-    // Call the function with each child
+    if (typeof node.children === "undefined") {
+        return resolve(true, '');
+    }
     var length = node.children.length;
     for (var i = 0; i < length; i++) {
-            validateNode(node.children[i], node.geometricError, resolve);
+        validateNode(node.children[i], node.geometricError, resolve);
     }
-
 }
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -23,14 +23,14 @@ function validateTileset(tileset) {
 function validateNode(root, parent, resolve) {
     var stack = [];
     stack.push({
-          node: root,
-          parent: parent
+        node: root,
+        parent: parent
     });
 
     while (stack.length > 0) {
         var node = stack.pop();
         var tile = node.node;
-        var parent = node.parent;
+        parent = node.parent;
 
         if (tile.geometricError > parent.geometricError) {
             return resolve({
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -1,0 +1,39 @@
+'use strict';
+var Promise = require('bluebird');
+
+module.exports = validateTileset;
+
+/**
+ * Checks whether the geometricError of each tile is <= the geometric error of its parent tile.
+ *
+ * @param {JSON} The JSON object representing the tileset.
+ * @return {Promise} A promise that resolves with two parameters - (1) a boolean for whether the tileset is valid
+ *                                                                 (2) the error message if the tileset is not valid.
+ *
+ */
+function validateTileset(tileset) {
+    if (tileset.root.geometricError > tileset.geometricError) {
+        return Promise.resolve(false, "Root has geometricError greater than tileset");
+    }
+
+    return new Promise(function(resolve, reject) {
+        checkGeometricError(tileset.root, tileset.geometricError, resolve);
+    });
+
+}
+
+function checkGeometricError(root, testError, resolve) {
+    // If we have reached a leaf, stop
+    if (root === undefined) {
+        return;
+    }
+    // Condition for invalid tileset
+    if (root.geometricError > testError) {
+        return resolve(false, "Child has geometricError greater than parent");
+    } else {
+        // Call the function with each child
+        for (var child in root.children) {
+            checkGeometricError(child, root.testError);
+        }
+    }
+}
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -12,9 +12,6 @@ module.exports = validateTileset;
  *
  */
 function validateTileset(tileset) {
-    if (tileset.root.geometricError > tileset.geometricError) {
-        return Promise.resolve(false, 'Root has geometricError greater than tileset');
-    }
 
     return new Promise(function(resolve) {
         validateNode(tileset.root, tileset.geometricError, resolve);
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -29,7 +29,10 @@ function validateNode(root, parentGeometricError, resolve) {
         if (typeof obj !== "undefined") {
 
             if (obj['node'].geometricError > obj['parentError']) {
-                return resolve(false, 'Child has geometricError greater than parent');
+                return resolve({
+                    result : false,
+                    message : 'Child has geometricError greater than parent'
+                });
             }
 
             if (typeof obj['node'].children !== "undefined") {
@@ -45,5 +48,8 @@ function validateNode(root, parentGeometricError, resolve) {
 
     }
 
-    return resolve(true, '');
+    return resolve({
+        result : true,
+        message : ''
+    });
 }
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var Promise = require('bluebird');
+var Cesium = require('Cesium');
+var defined = Cesium.defined;
 
 module.exports = validateTileset;
 
@@ -14,42 +16,44 @@ module.exports = validateTileset;
  */
 function validateTileset(tileset) {
     return new Promise(function(resolve) {
-        validateNode(tileset.root, tileset.geometricError, resolve);
+        validateNode(tileset.root, tileset, resolve);
     });
 }
 
-function validateNode(root, parentGeometricError, resolve) {
+function validateNode(root, parent, resolve) {
     var stack = [];
-    stack.push({'node': root, 'parentError': parentGeometricError});
+    stack.push({
+          node: root,
+          parent: parent
+        });
 
     while (stack.length > 0) {
 
-        var obj = stack.pop();
+        var node = stack.pop();
+        var tile = node.node;
+        var parent = node.parent;
 
-        if (typeof obj !== "undefined") {
+        if (tile.geometricError > parent.geometricError) {
+            return resolve({
+                result : false,
+                message : 'Child has geometricError greater than parent'
+            });
+        }
 
-            if (obj['node'].geometricError > obj['parentError']) {
-                return resolve({
-                    result : false,
-                    message : 'Child has geometricError greater than parent'
+        if (defined(tile.children)) {
+            var length = tile.children.length;
+            for (var i = 0; i < length; i++) {
+                stack.push({
+                    node: tile.children[i],
+                    parent: tile
                 });
             }
-
-            if (typeof obj['node'].children !== "undefined") {
-                var length = obj['node'].children.length;
-                for (var i = 0; i < length; i++) {
-                    if (typeof obj['node'].children[i] !== "undefined") {
-                        stack.push({'node': obj['node'].children[i], 'parentError': obj['node'].geometricError});
-                    }
-                }
-            }
-
         }
 
     }
 
     return resolve({
         result : true,
-        message : ''
+        message : 'Tileset is valid'
     });
 }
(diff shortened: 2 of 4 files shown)


validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+var Promise = require('bluebird');
+
+module.exports = validateTileset;
+
+/**
+ * Checks whether the geometricError of each tile is <= the geometric error of its parent tile.
```
lilleyse: Tweak this description so it is more generic, since `validateTileset` will do more than just check geometric error.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param {JSON} The JSON object representing the tileset.
```
lilleyse: Change `{JSON}` to `{Object}`

Also the doc needs to include the parameter name:

`@param {Object} tileset The JSON object representing the tileset.`

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+ * @return {Promise} A promise that resolves with two parameters - (1) a boolean for whether the tileset is valid
+ *                                                                 (2) the error message if the tileset is not valid.
+ *
+ */
+function validateTileset(tileset) {
+    if (tileset.root.geometricError > tileset.geometricError) {
```
lilleyse: Isn't this check handled automatically by the other function?

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+        return Promise.resolve(false, "Root has geometricError greater than tileset");
```
lilleyse: Typically we use single quotes rather than double quotes for strings.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
```
lilleyse: Remove this empty line.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+    return new Promise(function(resolve, reject) {
```
lilleyse: As we talked about offline, this approach is ok for now but may need some adjustments once we actually start to validate individual tiles. You can remove the `reject` parameter for now since it isn't used.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+        checkGeometricError(tileset.root, tileset.geometricError, resolve);
+    });
+
+}
+
+function checkGeometricError(root, testError, resolve) {
```
lilleyse: Rename `root` to `node`. Rename `testError` to `parentGeometricError`.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+    // If we have reached a leaf, stop
```
lilleyse: Since the code here is short and pretty self-explanatory, remove some of these comments.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+    if (root === undefined) {
+        return;
+    }
```
lilleyse: This block is not needed. I don't see when `root` would be undefined.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+    // Condition for invalid tileset
+    if (root.geometricError > testError) {
+        return resolve(false, "Child has geometricError greater than parent");
+    } else {
```
lilleyse: The `if` `else` isn't needed since the first path returns early.

```javascript
    if (root.geometricError > testError) {
        return resolve(false, "Child has geometricError greater than parent");
    }
    // Call the function with each child
    for (var child in root.children) {
        checkGeometricError(child, root.testError);
    }
```

is cleaner.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+        // Call the function with each child
+        for (var child in root.children) {
```
lilleyse: When iterating over an array we typically use just a basic for loop instead of for-in, to differentiate it from iterating over the keys of an object.

```javascript
var length = root.children.length;
for (var i = 0; i < length; ++i)
```

Also we typically factor out the length into its own variable. There are slight performance gains to doing it this way, and even though it won't make a noticeable difference in this code it is a good habit to get into.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+            checkGeometricError(child, root.testError);
```
lilleyse: Also pass the `resolve` parameter into here.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+}
```
lilleyse: Add newline.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+var validateTileset = require('../../lib/validateTileset');
```
lilleyse: Reverse the order of these. Third-party libraries typically are listed first.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+var clone = require('clone');
```
lilleyse: You need to `npm install clone --save`. In `package.json` remove the `^` that is added since we prefer to lock our library versions.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+
+describe('validateTileset', function() {
+
```
lilleyse: Remove empty line.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+    it('reads a valid JSON', function(done) {
+        var tileset = {
```
lilleyse: Place this variable outside the `it` because it will likely be used by multiple tests.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+                        },
+                        "geometricError": 0,
+                        "content": {
+                            "url": "c.b3dm"
+                        }
+                    }
+                ]
+            }
+        };
+        expect(validateTileset(tileset)).then(function(data) {
```
lilleyse: Style fix, put the `then` on the next line. In cesium we put it on the same line but have changed our style for our other projects.

```javascript
expect(validateTileset(result, errorMessage))
    .then(function(data) {
```

Also the function needs to take both parameters that are sent to it.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+           expect(data[0]).toBeTruthy();
```
lilleyse: Similar to above change `data[0]` to result. Change `toBeTruthy` to `toBe(true)`.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    it('reads an invalid JSON', function(done) {
+        // TODO
+    });
+
+    it('reads an invalid JSON', function(done) {
+        // TODO
+    });
+
```
lilleyse: Remove whitespace.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,66 @@
...
+});
```
lilleyse: Add newline.

validator/lib/validateTileset.js
lilleyse: Also rename `checkGeometricError` to `validateNode`. This tree traversal will eventually handle more than geometric error.

validator/specs/lib/validateTilesetSpec.js
lilleyse: The test should be modeled after `readTileSpec` to use the `done` and `toResolve` checks.

sumitshyamsukha: In this case, wouldn't we need to call `validateTileset` on `tileset` before we get `result` and `errorMessage`? I'm a little confused about how the above fragment actually calls `validateTileset` without passing in the `tileset` variable?

sumitshyamsukha: `expect(validateTileset(tileset))
            .then(function(result, errorMessage) {
           expect(result).toBe(true); 
        });`

I had something of this sort in mind.

lilleyse: Ah my bad, that code sample is definitely wrong. Should be:

```
validateTileset(tileset)
    .then(function(result, errorMesage) {
```

lilleyse: Your function is mostly good. Just look at `readTileSpec` for how to handle the `done` and `toResolve` checks. Otherwise if `validateTileset` were to not resolve the test would hang.

sumitshyamsukha:
@lilleyse On removing the undefined check, the tests seem to fail. I'll add that in again and recommit.

sumitshyamsukha:
@lilleyse The issue I'm running into now is in case of an invalid tileset, even though it resolves with an error message and a false value, the test seems to fail because the error message returned to `validateTileSpec` is `undefined`.

lilleyse:
I see your point with short circuiting the recursive function. Maybe recursion isn't flexible enough and we should do the iterative approach instead. Check out this example in Cesium: https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Cesium3DTileset.js#L1703-L1719

In the meantime I have a few minor comments about the tests.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,81 @@
...
+
+    it('reads a valid JSON', function(done) {
+        expect(validateTileset(tileset)
+            .then(function(result, errorMessage) {
+           expect(result).toBe(true);
+        }), done).toResolve();
+    });
+
+    var invalidTileset = clone(tileset);
+    invalidTileset.root.children[0].children[0].geometricError = 100;
```
lilleyse: These lines can go inside the `it`.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,81 @@
...
+    it('reads an invalid JSON', function(done) {
+        expect(validateTileset(invalidTileset)
+            .then(function(result, errorMessage) {
+                expect(result).toBe(false);
+                expect(errorMessage).toBe('Child has geometricError greater than parent');
+            }), done).toResolve();
+    });
+
+    invalidTileset = clone(tileset);
+    invalidTileset.root.geometricError = 300;
```
lilleyse: Here too.

sumitshyamsukha:
@lilleyse This seems to work, except the errorMessage for the invalid cases always ends up being undefined (but the truth value is correct). Is there something I'm doing incorrectly while resolving with the error messages? Please let me know if so, so I can fix it. The actual DFS logic seems to work fine.

lilleyse:
Maybe it doesn't like resolving with multiple arguments. You could try returning an object instead:

```javascript
return resolve({
    result : false,
    message : 'Child has geometricError greater than parent'
});
```

sumitshyamsukha:
Thank you @lilleyse 
That did the trick!

sumitshyamsukha:
@lilleyse Also added command line integration in the most recent commit. Let me know if it looks fine!

validator/bin/3d-tiles-validator.js
```diff
@@ -1,2 +1,15 @@
...
+var path = require('path');
+var Promise = require('bluebird');
+
+var readTileset = require('../lib/readTileset');
+var validateTileset = require('../lib/validateTileset');
+
+var filePath = argv._[0];
+var jsonTileset = readTileset(filePath);
+var response = validateTileset(jsonTileset);
+console.log(response);
```
lilleyse: `readTileset` and `validateTileset` are both promises, they don't return immediately.

Instead of just printing the response, print `response.message`.

Also add newline.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+function validateTileset(tileset) {
+    return new Promise(function(resolve) {
+        validateNode(tileset.root, tileset.geometricError, resolve);
+    });
+}
+
+function validateNode(root, parentGeometricError, resolve) {
+    var stack = [];
+    stack.push({'node': root, 'parentError': parentGeometricError});
+
```
lilleyse: Instead of adding the parent error, add the parent. This will be useful for validation besides geometric error.

Remove the quotes and format to multiple lines instead:

```javascript
stack.push({
    tile : root,
    parent : undefined
});
```

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+    while (stack.length > 0) {
+
+        var obj = stack.pop();
```
lilleyse: Create variables for `node` and `parent` here instead of accessing `obj['node']` and `obj['parentError']` all the time below. Also use dot notation for accessing members:

```javascript
var node = obj.node;
var parent = obj.parent;
```


For naming, it might be better if `obj` -> `node` and `node` -> `tile`.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+
+        if (typeof obj !== "undefined") {
```
lilleyse: Use Cesium's `defined` function instead. Though this should never be undefined so you can remove the line.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+                return resolve({
+                    result : false,
+                    message : 'Child has geometricError greater than parent'
+                });
+            }
+
+            if (typeof obj['node'].children !== "undefined") {
+                var length = obj['node'].children.length;
+                for (var i = 0; i < length; i++) {
+                    if (typeof obj['node'].children[i] !== "undefined") {
```
lilleyse: This should never be undefined.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+        if (typeof obj !== "undefined") {
+
+            if (obj['node'].geometricError > obj['parentError']) {
+                return resolve({
+                    result : false,
+                    message : 'Child has geometricError greater than parent'
+                });
+            }
+
+            if (typeof obj['node'].children !== "undefined") {
```
lilleyse: Like above, use `defined(tile.children)`.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+
+        }
+
+    }
+
+    return resolve({
+        result : true,
+        message : ''
```
lilleyse: Changing my mind here, but to keep things simple even the success could have a message like `Tileset is valid`.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,79 @@
...
+                    "geometricError": 0,
+                    "content": {
+                        "url": "c.b3dm"
+                    }
+                }
+            ]
+        }
+    };
+
+    it('reads a valid JSON', function(done) {
```
lilleyse: For these tests change `reads` to `validates`.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,79 @@
...
+        expect(validateTileset(tileset)
+            .then(function(response) {
+           expect(response.result).toBe(true);
+        }), done).toResolve();
+    });
```
lilleyse: The formatting is off for this test. The others look fine though.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,79 @@
...
+    it('reads an invalid JSON', function(done) {
+        var invalidTileset = clone(tileset);
+        invalidTileset.root.geometricError = 300;
+        expect(validateTileset(invalidTileset)
+            .then(function(response) {
+                expect(response.result).toBe(false);
+                expect(response.message).toBe('Child has geometricError greater than parent');
+            }), done).toResolve();
+    });
+
```
lilleyse: Remove empty line.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -0,0 +1,79 @@
...
+});
```
lilleyse: Add newline.

sumitshyamsukha:
@lilleyse Addressed all the comments, let me know if it's good to go!

validator/bin/3d-tiles-validator.js
```diff
@@ -1,2 +1,15 @@
...
+var argv = require('yargs').argv;
+var fsExtra = require('fs-extra');
+var path = require('path');
+var Promise = require('bluebird');
+
+var readTileset = require('../lib/readTileset');
+var validateTileset = require('../lib/validateTileset');
+
+var filePath = argv._[0];
+var response = validateTileset(readTileset(filePath));
```
lilleyse: This still needs to be fixed, using `.then`

Make sure to check that it works with a command like: `node bin/3d-tiles-validator.js specs/data/Tileset/tileset.json`

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,59 @@
...
+        validateNode(tileset.root, tileset, resolve);
+    });
+}
+
+function validateNode(root, parent, resolve) {
+    var stack = [];
+    stack.push({
+          node: root,
+          parent: parent
+        });
```
lilleyse: Needs slight formatting.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,59 @@
...
+
+    while (stack.length > 0) {
+
```
lilleyse: Remove empty line.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,59 @@
...
+        if (defined(tile.children)) {
+            var length = tile.children.length;
+            for (var i = 0; i < length; i++) {
+                stack.push({
+                    node: tile.children[i],
+                    parent: tile
+                });
+            }
+        }
+
```
lilleyse: Remove empty line.

validator/bin/3d-tiles-validator.js
```diff
@@ -1,2 +1,15 @@
 #!/usr/bin/env node
 'use strict';
+
+var argv = require('yargs').argv;
+var fsExtra = require('fs-extra');
+var path = require('path');
+var Promise = require('bluebird');
```
lilleyse: The last three aren't used, you can remove them.

validator/lib/validateTileset.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+var Promise = require('bluebird');
+var Cesium = require('Cesium');
```
lilleyse: Change to `cesium` (underscore).

sumitshyamsukha:
@lilleyse Should be fixed now -- tested the command line tool with the sample tileset.

lilleyse:
Made some tweaks. Thanks @sumitshyamsukha! This PR will be the foundation for a lot of the validation to come.

PR #26, opened 2016-11-02
Changed validateB3dm to return an object

JudyWeng:


lilleyse:
Perfect! Thanks @JudyWeng.

validator/lib/validateB3dm.js
```diff
@@ -27,17 +27,29 @@ function validateB3dm(content) {
     var byteLength = content.readUInt32LE(8);
 
     if (magic !== 'b3dm') {
-        return false;
+        return {
+            result : false,
```
lilleyse: > I also wonder if the object is overkill and if just returning a known string, e.g., 'VALID', is OK.

Maybe... it may be useful to have different success messages (or not..?). Since this change affects some of the open PRs and also @sumitshyamsukha's changes I think we should wait until those are in and then we can fix them all at once.

PR #29, opened 2016-11-08
Validate Bounding Volume

sumitshyamsukha:
@lilleyse This is just the start, will add more as we go


validator/lib/validateTileset.js
```diff
@@ -32,6 +32,19 @@ function validateNode(root, parent, resolve) {
...
+        if (defined(tile.content.boundingVolume)) {
+            var region = tile.content.boundingVolume.region;
+            var parentRegion = parent.content.boundingVolume.region;
+            for (var i = 0; i < region.length; i++) {
+                if (region[i] > parentRegion[i]) {
+                    return resolve({
+                        result: false,
+                        message: 'Child occupies region greater than parent'
+                    });
+                }
```
lilleyse: This check needs to be tweaked. A region is a longitude/latitude rectangle. The test needs to be more like a rectangle inside rectangle test (including also checking the heights). You should be able to use Cesium's `Rectangle` class to help out here. Each region can be converted to a rectangle, and then you can use `Rectangle.contains` on each corner. I would factor out this check into a function rather than in the tree traversal. In terms of contributing back to Cesium, this function may eventually become `Rectangle.containsRectangle`.

validator/lib/validateTileset.js
```diff
@@ -32,6 +32,19 @@ function validateNode(root, parent, resolve) {
         var tile = node.node;
         var parent = node.parent;
 
+        if (defined(tile.content.boundingVolume)) {
+            var region = tile.content.boundingVolume.region;
+            var parentRegion = parent.content.boundingVolume.region;
```
lilleyse: For this PR we aren't checking against the parent yet, instead check against `tile.boundingVolume`.

validator/lib/validateTileset.js
```diff
@@ -32,6 +32,19 @@ function validateNode(root, parent, resolve) {
         var tile = node.node;
         var parent = node.parent;
 
+        if (defined(tile.content.boundingVolume)) {
+            var region = tile.content.boundingVolume.region;
```
lilleyse: Check that `tile.content.boundingVolume.region` and `tile.boundingVolume.region` are defined before doing a region vs region check.

validator/bin/3d-tiles-validator.js
```diff
@@ -13,4 +13,4 @@ readTileset(filePath)
            .then(function(response) {
             console.log(response.message);
         })
-    });
+    });
```
lilleyse: Keep the newline in.

validator/lib/validateTileset.js
```diff
@@ -32,6 +32,19 @@ function validateNode(root, parent, resolve) {
         var tile = node.node;
         var parent = node.parent;
 
+        if (defined(tile.content.boundingVolume)) {
```
lilleyse: Also check that `tile.content` is defined. Empty tiles do not have content.

validator/lib/validateTileset.js
sumitshyamsukha: How is `tile.boundingVolume` different from `tile.content.boundingVolume`?

lilleyse: bv

In this screenshot with a parent and four children, `parent.boundingVolume` is white, `parent.content.boundingVolume` is blue, `tile.boundingVolume` are red, and `tile.content.boundingVolume` don't exist.

sumitshyamsukha: This makes a lot of sense. Thanks!

sumitshyamsukha:
@lilleyse Changed it to use scratch variables, please let me know if the logic seems correct.

sumitshyamsukha:
@lilleyse Added sphere inside sphere check + some refactoring

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
```
lilleyse: As mentioned offline, eventually this file should be merged into `validateTileset`.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+
+function regionInsideRegion() {
+    return (Rectangle.contains(scratchTileRect, Cartographic.fromRadians(scratchRect.west,
+        scratchRect.north, 0.0, scratchCartographic))
+    && Rectangle.contains(scratchTileRect, Cartographic.fromRadians(scratchRect.west,
+        scratchRect.south, 0.0, scratchCartographic))
+    && Rectangle.contains(scratchTileRect, Cartographic.fromRadians(scratchRect.east,
+        scratchRect.north, 0.0, scratchCartographic))
+    && Rectangle.contains(scratchTileRect, Cartographic.fromRadians(scratchRect.east,
+        scratchRect.south, 0.0, scratchCartographic)));
```
lilleyse: Just noticed.. for these you can use `Rectangle.southwest` and similar functions.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+  var sphereZ = sphere[2];
+  var sphereRadius = sphere[3];
+  var tileSphereX = tileSphere[0];
+  var tileSphereY = tileSphere[1];
+  var tileSphereZ = tileSphere[2];
+  var distance = (sphereX - tileSphereX)*(sphereX - tileSphereX) + (sphereY - tileSphereY)*(sphereY - tileSphereY) +
+      (sphereZ - tileSphereZ)*(sphereZ - tileSphereZ);
+  var tileRadius = tileSphere[3];
+
+  return (distance <= sphereRadius^2 && tileRadius <= sphereRadius);
```
lilleyse: In this function use `Cartesian3.unpack` and `CesiumMath.distance`.

Shouldn't it be `sphereRadius <= tileRadius`? The content bounding sphere is always contained within the tile's bounding sphere. Put parantheses around this statement so it is clearer.

`distance <= sphereRadius^2` - If you are doing `sphereRadius` squared, `sphereRadius ^ 2` is not what you want. I don't think that check is correct though, the checks needs to use both radius's in it.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+        parent: parent
+    });
+
+    while (stack.length > 0) {
+        var node = stack.pop();
+        var tile = node.node;
+        var parent = node.parent;
+
+        if (defined(tile.content)) {
+            if (defined(tile.content.boundingVolume)) {
```
lilleyse: These two lines can be combined. Factor out `tile.content` to its own variable to avoid repeating it a lot.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+                if (defined(tile.content.boundingVolume.region) && defined(tile.boundingVolume.region)) {
+                    var region = tile.content.boundingVolume.region;
+                    var tileRegion = tile.boundingVolume.region;
```
lilleyse: Rename for clarity. `tileRegion` is fine, but change `region` to `contentRegion`. Same applies elsewhere in this file.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+
+                    Cesium.Rectangle.fromRadians(region[0], region[1], region[2], region[3], scratchRect);
+                    Cesium.Rectangle.fromRadians(tileRegion[0], tileRegion[1], tileRegion[2], tileRegion[3],
+                                                scratchTileRect);
```
lilleyse: Add `var Rectangle = Cesium.Rectangle;` underneath `var defined = Cesium.defined;` at the top instead of accessing from Cesium like this.

Use `Rectangle.unpack`.

Instead of referencing `scratchRect` and `scratchTileRect` everywhere, creates new vars them like:

```javascript
var contentRectangle = Cesium.Rectangle.unpack(...)
var tileRectangle = Cesium.Rectangle.unpack(...)
```

And pass these along to `regionInsideRegion`.

Also, keeping the code on the same line is fine if it goes slightly over.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+
+                    var maxRectHeight = region[5];
+                    var maxTileHeight = tileRegion[5];
+
+                    if (!regionInsideRegion() || maxRectHeight > maxTileHeight) {
```
lilleyse: Also check the min heights.

lilleyse: Wrap `maxRectHeight > maxTileHeight` in parentheses like above.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+  var sphereRadius = sphere[3];
+  var tileSphereX = tileSphere[0];
+  var tileSphereY = tileSphere[1];
+  var tileSphereZ = tileSphere[2];
+  var distance = (sphereX - tileSphereX)*(sphereX - tileSphereX) + (sphereY - tileSphereY)*(sphereY - tileSphereY) +
+      (sphereZ - tileSphereZ)*(sphereZ - tileSphereZ);
+  var tileRadius = tileSphere[3];
+
+  return (distance <= sphereRadius^2 && tileRadius <= sphereRadius);
+
```
lilleyse: Remove whitespace.

lilleyse: The indentation in this function should be 4 spaces rather than 2.

sumitshyamsukha:
@lilleyse Incorporated your comments and added a region inside region test, will add a sphere inside sphere test to this too. Once everything works, I'll refactor the `validateBoundingVolume` file back into `validateTileset` if that's okay with you.

pjcozzi:
@sumitshyamsukha can you please merge in master?  There is probably a trivial conflict to resolve.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+        contentRect.north, 0.0, scratchCartographic))
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.west,
+        contentRect.south, 0.0, scratchCartographic))
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.east,
+        contentRect.north, 0.0, scratchCartographic))
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.east,
+        contentRect.south, 0.0, scratchCartographic)));
+}
+
+function sphereInsideSphere(contentSphere, tileSphere) {
```
lilleyse: Yeah that's the end goal. Our plan was to write them here first and then push them over. Ideally we should get these into core Cesium before 12/1.

pjcozzi: OK, please document in an issue/roadmap.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+
+function validateVolume(tileset) {
+    return new Promise(function(resolve) {
+        validateNode(tileset.root, tileset, resolve);
+    });
+}
+
+var scratchCartographic = new Cartographic();
+
+function regionInsideRegion(contentRect, tileRect) {
```
lilleyse: Using the full names `contentRectangle` and `tileRectangle` is preferable.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.west,
+        contentRect.south, 0.0, scratchCartographic))
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.east,
+        contentRect.north, 0.0, scratchCartographic))
+    && Rectangle.contains(tileRect, Cartographic.fromRadians(contentRect.east,
+        contentRect.south, 0.0, scratchCartographic)));
+}
+
+function sphereInsideSphere(contentSphere, tileSphere) {
+    var contentSphereRadius = contentSphere[3];
```
lilleyse: Just `contentRadius` is fine.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+    var tileRadius = tileSphere[3];
+    var distance = CesiumMath.distance(Cartesian3.unpack(contentSphere, 0), Cartesian3.unpack(tileSphere, 0));
```
lilleyse: Create `var contentCenter` and `var tileCenter` instead of including in one line. Also you can use two scratch Cartesian3 for the unpack method.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+    if (Math.min(contentSphereRadius, tileRadius) == contentSphereRadius) {
+        return ((distance + contentSphereRadius) <= tileRadius);
+    } else {
+        return false;
+    }
```
lilleyse: This can be condensed into one line. You don't need the if check - if the content radius is bigger it will still satisfy the equation correctly.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+        var parent = node.parent;
+        var tileContent = tile.content;
+
+        if (defined(tileContent) && defined(tileContent.boundingVolume)) {
+            if (defined(tileContent.boundingVolume.region) && defined(tile.boundingVolume.region)) {
+                var contentRegion = tileContent.boundingVolume.region;
+                var tileRegion = tile.boundingVolume.region;
+
+                var contentRect = Rectangle.unpack(contentRegion, 0);
+                var tileRect = Rectangle.unpack(tileRegion, 0);
```
lilleyse: You can still use scratch variables here, like:

`var contentRect = Rectangle.unpack(contentRegion, 0, scratchContentRectangle);`

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+
+                var maxRectHeight = contentRegion[5];
```
lilleyse: Rename to `maxContentHeight` to be consistent with tile/content naming scheme.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+                var maxTileHeight = tileRegion[5];
+                var minRectHeight = contentRegion[4];
+                var minTileHeight = tileRegion[4];
+
+                if (!regionInsideRegion(contentRect, tileRect) || (maxRectHeight > maxTileHeight) || (minRectHeight > minTileHeight)) {
```
lilleyse: Should be `minRectHeight < minTileHeight`

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,102 @@
...
+                    return resolve({
+                        result: false,
+                        message: 'Child bounding volume is not contained within parent'
+                    });
+                }
+            }
+
+            if (defined(tileContent.boundingVolume.sphere) && defined(tile.boundingVolume.sphere)) {
+                var sphere = tileContent.boundingVolume.sphere;
```
lilleyse: Rename to `contentSphere`.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,60 @@ describe('validateTileset', function() {
...
+                       "region": [-1.31968, 0.6988424218, -1.3196390408203893, 0.698874, 0, 20]
+                   },
+                   "geometricError": 0,
+                   "content": {
+                       "url": "c.b3dm"
+                   }
+               }
+           ]
+       }
+   };
```
lilleyse: For testing bounding volumes the sample tileset can be simpler and just cover the needed cases. You could even get away with just a root node for now that has a boundingVolume and contentBoundingVolume.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,60 @@ describe('validateTileset', function() {
...
+
+    it('validates region inside region', function(done) {
+        expect(validateBoundingVolume(tileset)
+            .then(function(response) {
+                expect(response.result).toBe(true);
+                expect(response.message).toBe('Tileset is valid');
+            }), done).toResolve();
+    });
```
lilleyse: Along with these test, add some failure cases like:
* Content region max height is taller than tile region
* Content region min height is smaller than tile region
* Individually test cases where the content region's west/east/south/north values exceed the tile region

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,101 @@
...
+        && Rectangle.contains(tileRectangle, Rectangle.southeast(contentRectangle, scratchCartographic)));
+}
+
+function sphereInsideSphere(contentSphere, tileSphere) {
+    var contentRadius = contentSphere[3];
+    var tileRadius = tileSphere[3];
+    var contentCenter = Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);
+    var tileCenter = Cartesian3.unpack(tileSphere, 0, scratchTileCartesian);
+    var distance = Cartesian3.distance(contentCenter, tileCenter);
+    return ((distance + contentRadius) <= tileRadius);
```
lilleyse: This may be more clear as `distance <= (tileRadius - contentRadius)`

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,101 @@
...
+        var parent = node.parent;
+        var tileContent = tile.content;
+
+        if (defined(tileContent) && defined(tileContent.boundingVolume)) {
+            if (defined(tileContent.boundingVolume.region) && defined(tile.boundingVolume.region)) {
+                var contentRegion = tileContent.boundingVolume.region;
+                var tileRegion = tile.boundingVolume.region;
+
+                var contentRect = Rectangle.unpack(contentRegion, 0, scratchContentRectangle);
+                var tileRect = Rectangle.unpack(tileRegion, 0, scratchTileRectangle);
```
lilleyse: Name these `contentRectangle` and `tileRectangle` instead.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,101 @@
...
+
+                var maxContentHeight = contentRegion[5];
+                var minContentHeight = contentRegion[4];
+                var maxTileHeight = tileRegion[5];
+                var minTileHeight = tileRegion[4];
+
+                if (!regionInsideRegion(contentRect, tileRect) || (maxContentHeight > maxTileHeight) || (minContentHeight < minTileHeight)) {
+                    return resolve({
```
lilleyse: Similar to how `sphereInsideSphere` unpacks inside its function, you could also bring these lines into `regionInsideRegion` and pass in `contentRegion` and `tileRegion`. Then the rectangle check in `regionInsideRegion` can be factored out into a function called `rectangleInsideRectangle`.

validator/lib/validateBoundingVolume.js
lilleyse: For your next update it would good to do this merge.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,98 @@ describe('validateTileset', function() {
...
+
+describe('validateBoundingVolume', function() {
+   var regionTileset = {
+       "root": {
+           "boundingVolume": {
+               "region": [10, 10, 10, 10, 10, 88]
+           },
+           "content": {
+               "boundingVolume": {
+                   "region":  [0, 0, 0, 0, 20, 30]
```
lilleyse: It would make sense to create this tileset so it passes. Right now the content region isn't inside the tile region.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,98 @@ describe('validateTileset', function() {
...
+               },
```
lilleyse: Remove comma.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,98 @@ describe('validateTileset', function() {
...
+           }
+       }
+   };
```
lilleyse: The spacing for this function is slightly off.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,98 @@ describe('validateTileset', function() {
...
+    it('validates an invalid region inside region', function(done) {
+        var invalidTileset = clone(regionTileset);
+        expect(validateBoundingVolume(invalidTileset)
+            .then(function(response) {
+                expect(response.result).toBe(false);
+                expect(response.message).toBe('Child bounding volume is not contained within parent');
+            }), done).toResolve();
+    });
+
+    it('validates content region max height taller than tile region', function(done) {
```
lilleyse: Reword so it is clear that it is checking this as invalid. Same below.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,98 @@ describe('validateTileset', function() {
...
+
+    it('validates region inside region', function(done) {
+        expect(validateBoundingVolume(regionTileset)
+            .then(function(response) {
+                expect(response.result).toBe(true);
+                expect(response.message).toBe('Tileset is valid');
+            }), done).toResolve();
+    });
+
+    it('validates an invalid region inside region', function(done) {
```
lilleyse: Slightly awkward wording. Maybe `Invalidates when a content region exceeds the tile region`.

validator/specs/lib/validateTilesetSpec.js
sumitshyamsukha: My understanding of this might be slightly incorrect, but I updated it to be a tileset that I thought should work, namely:
`content: [18, 14, 15, 8]` and `tile: [20, 40, 50, 10]`

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -81,18 +81,18 @@ describe('validateTileset', function() {
...
        "root": {
            "boundingVolume": {
-               "region": [10, 10, 10, 10, 10, 88]
+               "region": [20, 40, 50, 10, 10, 88]
            },
            "content": {
                "boundingVolume": {
-                   "region":  [0, 0, 0, 0, 20, 30]
-               },
+                   "region":  [18, 14, 15, 8, 20, 30]
```
lilleyse: > content: [18, 14, 15, 8] and tile: [20, 40, 50, 10]

Based on the layout: [west, south, east, north]

* Within the content region or tile region, west should be less than east, and south should be less north.
* The content west/east don't fit within the tile's west/east
* The content south/north also don't fit.

sumitshyamsukha: Thanks so much! Fixed in next commit.

sumitshyamsukha:
@lilleyse Added sphere inside region function with a couple of tests, although I'm not sure how to check if the sphere's center is contained within the region -- any suggestions for that? Thanks!

lilleyse:
To check the center, convert the sphere's center to a `Cartographic`. For this use `Cartographic.fromCartesian` function. Then just check if it is contained in the rectangle and between the heights like normal.

sumitshyamsukha:
@lilleyse Added the check for the center, please let me know if there's any other issues.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,123 @@
...
+    var maxContentHeight = contentRegion[5];
+    var minContentHeight = contentRegion[4];
+    var maxTileHeight = tileRegion[5];
+    var minTileHeight = tileRegion[4];
+    return (Rectangle.contains(tileRectangle,  Rectangle.northwest(contentRectangle, scratchCartographic))
+        && Rectangle.contains(tileRectangle, Rectangle.southwest(contentRectangle, scratchCartographic))
+        && Rectangle.contains(tileRectangle, Rectangle.northeast(contentRectangle, scratchCartographic))
+        && Rectangle.contains(tileRectangle, Rectangle.southeast(contentRectangle, scratchCartographic)))
+        && (maxContentHeight < maxTileHeight)
+        && (minContentHeight > minTileHeight);
```
lilleyse: Use <= and >=

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,123 @@
...
+    var contentCenter = Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);
+    var tileCenter = Cartesian3.unpack(tileSphere, 0, scratchTileCartesian);
+    var distance = Cartesian3.distance(contentCenter, tileCenter);
+    return distance  <= (tileRadius - contentRadius);
+}
+
+function sphereInsideRegion(contentSphere, tileRegion) {
+    var contentRadius = contentSphere[3];
+    Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);
+    Cartographic.fromCartesian(scratchContentCartesian, Cesium.Ellipsoid.WGS84, scratchCartographic);
```
lilleyse: Include `Ellispoid` above with the other Cesium includes.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,123 @@
...
+        && (minContentHeight > minTileHeight);
+}
+
+function sphereInsideSphere(contentSphere, tileSphere) {
+    var contentRadius = contentSphere[3];
+    var tileRadius = tileSphere[3];
+    var contentCenter = Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);
+    var tileCenter = Cartesian3.unpack(tileSphere, 0, scratchTileCartesian);
+    var distance = Cartesian3.distance(contentCenter, tileCenter);
+    return distance  <= (tileRadius - contentRadius);
```
lilleyse: Remove extra space between `distance` and `<=`.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function sphereInsideRegion(contentSphere, tileRegion) {
+    var contentRadius = contentSphere[3];
+    Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);
```
lilleyse: Like in `sphereInsideSphere` create a local var for this instead of referencing the scratch variable multiple times.

`var contentCenter = Cartesian3.unpack(contentSphere, 0, scratchContentCartesian);`

lilleyse: As well as for the line below, `var contentCartographic` maybe?

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,123 @@
...
+    Cartographic.fromCartesian(scratchContentCartesian, Cesium.Ellipsoid.WGS84, scratchCartographic);
+    var tileRectangle = Rectangle.unpack(tileRegion, 0, scratchTileRectangle);
+    var width = tileRegion[2] - tileRegion[0];
+    var height = tileRegion[3] - tileRegion[1];
+    return ((Rectangle.contains(tileRectangle, scratchCartographic)) &&
+            (2*contentRadius <= width && 2*contentRadius <= height));
+}
```
lilleyse: There are a couple of problems with this test, it doesn't take the tile region's min and max height into account and doesn't account for when the sphere is near the edge of the region.

I haven't fully thought this through, but if you get the normals of each side of the region, you can use those to get 6 points on the sphere. Then check that each point is inside the region. If may be helpful to write a `cartographicInsideRegion` function used here and in `regionInsideRegion`.

This function may be helpful for finding the normals: https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/TileBoundingRegion.js#L160

lilleyse:
Also make sure to resolve the merge conflict and fix jsHint errors.

validator/lib/validateBoundingVolume.js
sumitshyamsukha: I tried doing this using `contentCartographic` and rerunning the tests, but the promise doesn't resolve giving me the error that a cartographic is required. However, when I change it back to `scratchCartographic` it works fine. Any suggestions on why this might be happening?

sumitshyamsukha:
@lilleyse Removed sphere inside region test for now, let me know if it's good to merge

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,160 @@ describe('validateTileset', function() {
...
+        var invalidSphere = clone(sphereTileset);
+        invalidSphere.root.content.boundingVolume.sphere[2] = 6;
+        expect(validateBoundingVolume(invalidSphere)
+            .then(function(response) {
+                expect(response.result).toBe(false);
+                expect(response.message).toBe('Child bounding volume is not contained within parent');
+            }), done).toResolve();
+    });
+
+    var hybridTileset = {
```
lilleyse: Remove everything under here now that sphere inside region is removed.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
+
+var Promise = require('bluebird');
```
lilleyse: Make sure to merge the code in `validateBoundingVolume` into `validateTileset`.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
...
+            if (defined(contentSphere) && defined(tileSphere)) {
+                if (!sphereInsideSphere(contentSphere, tileSphere)) {
+                    return resolve({
+                        result: false,
+                        message: 'Child bounding volume is not contained within parent'
+                    });
+                }
+            }
+
+            if (defined(contentSphere) && defined(tileRegion)) {
```
lilleyse: Remove this check now.

validator/lib/validateBoundingVolume.js
```diff
@@ -0,0 +1,109 @@
+'use strict';
+
+var Promise = require('bluebird');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var Cartesian3 = Cesium.Cartesian3;
+var Rectangle = Cesium.Rectangle;
+var Cartographic = Cesium.Cartographic;
+var Ellipsoid = Cesium.Ellipsoid;
```
lilleyse: Remove `Ellipsoid`.

sumitshyamsukha:
@lilleyse Merged the bounding volume checks into `validateTileset`

lilleyse:
Thanks, just delete `validateBoundingVolume.js` and this should be good to go.

sumitshyamsukha:
@lilleyse Done!

lilleyse:
Thanks @sumitshyamsukha!

PR #27, opened 2016-11-02
Validate i3dm header

JudyWeng:
#22 
I've ran jsDoc, jsHint, and the tests and everything seems fine so far. I usually test every unit test possible, and therefore made 2 valid i2dm tiles with the different gITF formats. Please let me know if I should just keep it to 1 valid tile instead or if anything else needs to be changed.

Issue #22, opened 2016-10-28
Validator - validate i3dm header


lilleyse:
@JudyWeng

Like https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/16, but for i3dm tiles.

Check that the header is valid according to the spec. Ignore the feature table and batch table for now, but make sure to check that `gltfFormat` is valid.

https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Instanced3DModel

PR #31, opened 2016-11-12
Validate i3dm new

JudyWeng:
Addressing Issue #22, continuing from #27 due to rebasing multiple commits from other merges into validator-master


lilleyse:
Thanks @JudyWeng all looks good to me.

Issue #23, opened 2016-10-28
Validator - validate cmpt header


lilleyse:
Validate composite tiles (cmpt) according to the spec. https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Composite

A composite tile contains other tiles (possibly even other composite tiles). You will need to extract each of these tiles and run them through their respective validator functions, and only return true if all of them pass. You can check out the Cesium function for ideas on how to extract the interior tiles: https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Composite3DTileContent.js#L174-L235.

lilleyse:
@JudyWeng

PR #33, opened 2016-11-19
Validate cmpt header

JudyWeng:
#23 
Please let me know if there is anything that I should change or that I missed.

### DIFF
--- a/validator/lib/validateCmpt.js
+++ b/validator/lib/validateCmpt.js
@@ -25,7 +25,9 @@ function validateCmpt(content) {
         throw new DeveloperError('content must be of type buffer');
     }
 
-    if (content.length < 16) {
+    var headerByteLength = 16;
+
+    if (content.length < headerByteLength) {
         return {
             result : false,
             message: 'Cmpt header must have min byte length of 16. Current header length = ' + content.length
@@ -63,7 +65,6 @@ function validateCmpt(content) {
     }
 
     for (var i = 0; i < tilesLength; i++) {
-
         var errorAddon = '';
         var validatorResult;
 
@@ -76,14 +77,15 @@ function validateCmpt(content) {
 
         var innerTileMagic = content.toString('utf8', byteOffset, byteOffset + sizeOfUint32);
         var innerTileByteLength = content.readUInt32LE(byteOffset + 2 * sizeOfUint32);
-        var innerTile = content.slice(byteOffset, byteOffset + innerTileByteLength);
 
         if (byteOffset + innerTileByteLength > byteLength) {
             return {
                 result : false,
-                message: 'Inner tile exceeds provided buffer\'s length. Byte length = ' + byteLength + '. Inner tile\'s end = ' + byteOffset + innerTileByteLength
+                message: 'Inner ' + innerTileMagic + ' tile exceeds provided buffer\'s length. Byte length = ' + byteLength + '. Inner tile\'s end = ' + byteOffset + innerTileByteLength
             };
         }
+
+        var innerTile = content.slice(byteOffset, byteOffset + innerTileByteLength);
 
         if (innerTileMagic === 'b3dm') {
             validatorResult = validateB3dm(innerTile);
@@ -103,7 +105,7 @@ function validateCmpt(content) {
             errorAddon += validatorResult.message;
         } else {
             isValid = false;
-            errorAddon += '\nInner tile header magic cannot be identified; header = ' + innerTileMagic;
+            errorAddon += 'Inner tile header magic cannot be identified; header = ' + innerTileMagic;
         }
 
         byteOffset = byteOffset + innerTileByteLength; // skip over this tile
@@ -111,7 +113,7 @@ function validateCmpt(content) {
         if (!isValid) {
             var errorMessage = 'Cmpt header has an invalid inner tile:\n';
             errorMessage += 'Invalid inner tile index = ' + i + ' starting at byte = ' + (byteOffset - innerTileByteLength) + '\n';
-            errorMessage += 'Invalid inner tile has magic = ' + innerTileMagic;
+            errorMessage += 'Invalid inner tile has magic = ' + innerTileMagic + '\n';
             errorMessage += errorAddon;
 
             return {
@@ -122,7 +124,7 @@ function validateCmpt(content) {
     }
 
     return {
-        result : isValid,
+        result : true,
         message: 'valid cmpt'
     };
 }
--- a/validator/lib/validateCmpt.js
+++ b/validator/lib/validateCmpt.js
@@ -1,6 +1,6 @@
 'use strict';
 var Cesium = require('cesium');
-var validateB3dm = require('../lib/validateB3dm.js');
+var validateB3dm = require('../lib/validateB3dm');
 var validateI3dm = require('../lib/validateI3dm');
 var validatePnts = require('../lib/validatePnts');
 
@@ -23,6 +23,13 @@ function validateCmpt(content) {
 
     if (!Buffer.isBuffer(content)) {
         throw new DeveloperError('content must be of type buffer');
+    }
+
+    if (content.length < 16) {
+        return {
+            result : false,
+            message: 'Cmpt header must have min byte length of 16. Current header length = ' + content.length
+        };
     }
 
     var byteOffset = 16;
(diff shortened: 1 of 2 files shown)


validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
+'use strict';
+var Cesium = require('cesium');
+var validateB3dm = require('../lib/validateB3dm.js');
```
lilleyse: Remove `.js`

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
...
+        var errorAddon = '';
+        var innerTileMagic = content.toString('utf8', byteOffset, byteOffset + sizeOfUint32);
+        var innerTileByteLength = content.readUInt32LE(byteOffset + 2 * sizeOfUint32);
+
+        var innerTile;
+        var validatorResult;
+
+        if (innerTileMagic === 'b3dm') {
+            innerTile = new Buffer(24);
+            content.copy(innerTile, 0, byteOffset, byteOffset + innerTileByteLength);
```
lilleyse: Use `Buffer.slice` instead for these.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
...
+            validatorResult = validateB3dm(innerTile);
+            isValid = isValid && validatorResult.result;
+        }
+
```
lilleyse: Remove the empty spaces between each of these.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
...
+        else if (innerTileMagic === 'cmpt') {
+            innerTile = new Buffer(innerTileByteLength);
+            content.copy(innerTile, 0, byteOffset, byteOffset + innerTileByteLength);
+            validatorResult = validateCmpt(innerTile);
+            isValid = isValid && validatorResult.result;
+        }
+
+        else {
+            isValid = false;
+            errorAddon += 'Inner tile header magic cannot be identified';
```
lilleyse: Include the `innerTileMagic` in the error message.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        byteOffset = byteOffset + innerTileByteLength; // skip over this tile
+
+        if( (byteOffset + 12 > byteLength) && (i + 1 !== tilesLength) ) {
+            isValid = false;
+            errorAddon += 'Cmpt header given tilesLength = ' + tilesLength + '. Found number of inner tiles = ' + (i + 1);
+        }
```
lilleyse: This check and the one at the end


```javascript
    if (byteOffset !== byteLength) {
        return {
            result : false,
            message: 'Header has invalid inner tile formats or tileLength field. Expected end of tile = ' + byteLength + '. Found end of tile = ' + byteOffset
        };
    }
```

can be combined together. And a different test can be added to differentiate between the tile byte length going too far vs. an inner tile not existing.

It will be clearer if all this is near the beginning of the loop. The message can also be simpler, just that an inner tile's byte length exceeds the cmpt tile's byte length. Before this check you can have another check that sees if `byteOffset + 3 * sizeOfUint32` exceeds `byteLength` (aka the buffer doesn't actually contain this tile and can't even get the tile's byte length). If this is true than the error message can be the one you have here, that the cmpt tile does not have as many inner tiles as it says it has.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,136 @@
...
+
+        if (!isValid) {
+            var errorMessage = 'Cmpt header has an invalid inner tile:\n';
+            errorMessage += 'Invalid inner tile index = ' + i + ' starting at byte = ' + (byteOffset - innerTileByteLength) + '\n';
+            errorMessage += 'Invalid inner tile has magic = ' + innerTileMagic;
+            if (defined(validatorResult)) {
+                errorMessage += '\n' + validatorResult.message;
+            }
+
+            if(errorAddon.length !== 0) {
```
lilleyse: `if (`

Instead of having both `validatorResult.message` and `errorAddon` you can combine these into the same variable. Then the code becomes shorter and simpler:

```javascript
errorMessage += 'Invalid inner tile has magic = ' + innerTileMagic + '\n';
errorMessage += errorAddon;
```

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,495 @@
...
+
+    //cmpt header
+    cmptTile.write('cmpt', byteOffset); // magic
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(1, byteOffset); // version
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(cmptTile.length, byteOffset); // byteLength
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(5, byteOffset); // tilesLength
+    byteOffset += sizeU32Int;
```
lilleyse: Instead of going with the `byteOffset` approach for building the test data, it may be worth considering `Buffer.concat` where the cmpt and i3dm and created separately and then merged (wherever this makes sense).

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,495 @@
...
+    cmptTile.writeUInt32LE(1, byteOffset); // version
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(cmptTile.length, byteOffset); // byteLength
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(1, byteOffset); // tilesLength
+    byteOffset += sizeU32Int;
+
+    //unknown header
+    cmptTile.write('xxxx', byteOffset); // magic
```
lilleyse: Provide valid data for the rest of the header so it doesn't trigger a different throw.

validator/lib/validateCmpt.js
pjcozzi: `else` can actually go on the same line as `}` like the Cesium code.

pjcozzi:
@JudyWeng nice job with this, especially covering all those cases.

validator/lib/validateCmpt.js
JudyWeng: The check 
```
if (byteOffset !== byteLength) {
        return {
            result : false,
            message: 'Header has invalid inner tile formats or tileLength field. Expected end of tile = ' + byteLength + '. Found end of tile = ' + byteOffset
        };
    }
```
checks the condition where all inner tiles are valid, but there are extra bytes left over. I didn't see a more readable way to combine it with the other check indicated above. I did move the other check, but the only way I can see to properly check this is to check it after the for-loop finishes because that's when we know all inner tiles are valid.

lilleyse: Extra bytes leftover is actually ok. Sometimes this is actually necessary for padding purposes.

As long as you check that a tile's `byteLength` doesn't cause it to exceed the buffer length, that's good enough.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,128 @@
...
+function validateCmpt(content) {
+    if (!defined(content)) {
+        throw new DeveloperError('cmpt content must be defined');
+    }
+
+    if (!Buffer.isBuffer(content)) {
+        throw new DeveloperError('content must be of type buffer');
+    }
+
+    if (content.length < 16) {
```
lilleyse: Save `16` off as a variable called `headerByteLength`. This will make it easier to follow.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,128 @@
...
+
+    if (byteLength !== content.length) {
+        return {
+            result : false,
+            message: 'Cmpt header has an invalid byteLength field. Expected byteLength = ' + content.length + '. Found byteLength = ' + byteLength
+        };
+    }
+
+    for (var i = 0; i < tilesLength; i++) {
+
```
lilleyse: Remove whitespace.

validator/lib/validateCmpt.js
```diff
@@ -0,0 +1,128 @@
...
+            return {
+                result : false,
+                message: errorMessage
+            };
+        }
+    }
+
+    return {
+        result : isValid,
```
lilleyse: Since it will always be true, just use `result : true`.

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,279 @@
...
+    pntsTile.write('pnts', byteOffset); // magic
+    byteOffset += sizeU32Int;
+    pntsTile.writeUInt32LE(version, byteOffset); // version
+    byteOffset += sizeU32Int;
+    pntsTile.writeUInt32LE(byteLength, byteOffset); // byteLength
+
+    return pntsTile;
+}
+
+function createUnknownTile(version, byteLength) {
```
lilleyse: Instead of this, just call `createB3dmTile` and then edit the magic field.

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,279 @@
...
+    unknownTile.write('xxxx', byteOffset); // magic
+    byteOffset += sizeU32Int;
+    unknownTile.writeUInt32LE(version, byteOffset); // version
+    byteOffset += sizeU32Int;
+    unknownTile.writeUInt32LE(byteLength, byteOffset); // byteLength
+
+    return unknownTile;
+}
+
+function createShortHeader() {
```
lilleyse: This can also be replaced by calling `createCmpt` with an empty array of tile buffers, and then `buffer.slice(cmptHeaderSize - 4)`

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,279 @@
...
+    cmptTile.write('cmpt', byteOffset); // magic
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(1, byteOffset); // version
+    byteOffset += sizeU32Int;
+    cmptTile.writeUInt32LE(cmptTile.length, byteOffset); // byteLength
+
+    return cmptTile;
+}
+
+function createInvalidMagic() {
```
lilleyse: Similarly, call `createCmpt` with empty array and then edit the magic after.

Many of the functions below can follow a similar strategy - create a valid cmpt and then edit it after so it is invalid.

JudyWeng:
I've updated the test case spec file according to the comments. Please let me know if there is anything else I should change

pjcozzi:
@lilleyse are you able to review this?

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,191 @@
...
+function createB3dmTile() {
+    var b3dmTile = new Buffer(b3dmHeaderSize);
+    b3dmTile.write('b3dm', 0); // magic
+    b3dmTile.writeUInt32LE(1, 4); // version
+    b3dmTile.writeUInt32LE(b3dmHeaderSize, 8); // byteLength
+
+    return b3dmTile;
+}
+
+function createI3dmTile(gltfFormat) {
```
lilleyse: Better to follow the approach of always generating a valid `i3dm` tile here and then modifying the `gltfFormat` to be wrong later.

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,191 @@
...
+    i3dmTile.writeUInt32LE(gltfFormat, 28); // gltfFormat: 0 - url
+
+    return i3dmTile;
+}
+
+function createCmptTile(tiles) {
+    var byteLength = cmptHeaderSize;
+    for(var i = 0; i < tiles.length; i++) {
+        byteLength += tiles[i].readUInt32LE(8);
+    }
```
lilleyse: No need to find the byte length and send it to `Buffer.concat`. `Buffer.concat(tiles)` is fine.

validator/specs/lib/validateCmptSpec.js
```diff
@@ -0,0 +1,191 @@
...
+
+    var cmptTile = new Buffer(cmptHeaderSize);
```
lilleyse: Name this `cmptHeader` instead.

pjcozzi:
Any update here? @JudyWeng do you have the bandwidth to make the changes suggested in @lilleyse's latest review?

JudyWeng:
I can make the changes here. I just some things due this Wednesday that I was trying to finish up first. Sorry for the delay.

lilleyse:
Sorry for the delay, this looks good!

Issue #35, opened 2016-12-05
Validator - Validate tiles referenced in tileset.json


lilleyse:
@sumitshyamsukha 

When traversing the tileset.json, if the tile's content references a url, load that tile into memory, check its magic to determine its type, and then check if it is valid according to the `validateB3dm`, `validateI3dm`, `validatePnts`, and `validateCmpt` functions.

`validateCmpt` is not in yet but should be soon. You can model the functionality here after the code in there.

The trickiest part will be managing the promises. Let me know if you need help with that.

sumitshyamsukha:
@lilleyse How should I go about loading the tile into memory? Just use a variable or is there a better way? Also, should it be valid according to all of the 4 mentioned functions, or any one of them? Thanks!

lilleyse:
You can use the `readTile` function that will give you a Buffer object. Based on the magic in the tile's header, it will check only one of the 4 functions.

PR #37, opened 2016-12-06
Validate tiles referenced

sumitshyamsukha:
For #35 

@lilleyse Let me know if you think this is a reasonable way to go about it. Thanks!


validator/lib/validateTileset.js
```diff
@@ -32,6 +37,71 @@ function validateNode(root, parent, resolve) {
...
+                            result: false,
+                            message: 'Child occupies region greater than parent'
+                        });
+                    }
+                }
+            }
+        }
+
+        if (defined(tile.content) && defined(tile.content.url)) {
+            var tileBuffer = readTile(tile.content.url);
```
lilleyse: `readTile` returns a promise.

To handle multiple promises you can create an array at the top of `validateNode` and push the promises to this. Then at the end of the function you can use `Promise.all(tilePromises)` which inside it can return:

```
    return resolve({
        result : true,
        message : 'Tileset is valid'
    });
```

validator/lib/validateTileset.js
```diff
@@ -32,6 +37,71 @@ function validateNode(root, parent, resolve) {
...
+            if (defined(tileBuffer)) {
+                var magic = tileBuffer.toString('utf8', 0, 4);
+                if (magic === 'b3dm') {
+                    validateB3dm(tileBuffer)
```
lilleyse: The validate functions do not return promises

sumitshyamsukha:
@lilleyse Added support to handle multiple promises, could you tell me if this way is what you were hinting at?

sumitshyamsukha:
@lilleyse Added the `readTile` promise to the list of promises. Let me know if everything is good to go.

lilleyse:
You can remove the `tilePromises.push(resolve({` areas.

I may have more comments later - can you added tests and if possible remove the commits from https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/29? If that's not easily doable we should merge https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/29 first.

sumitshyamsukha:
@lilleyse I think it might be a better idea to merge #29 first... I'll try to finish that up

sumitshyamsukha:
@lilleyse The test seems to result in an async exception... not sure how to fix this: any ideas would be appreciated!

validator/bin/3d-tiles-validator.js
```diff
@@ -9,8 +9,8 @@ var validateTileset = require('../lib/validateTileset');
...
     .then(function(json) {
-        validateTileset(json)
-            .then(function(response) {
-                console.log(response.message);
-            });
+       validateTileset(json)
+           .then(function(response) {
+            console.log(response.message);
+        });
     });
```
lilleyse: The new formatting here is messy, the old spacing seemed to be fine.

validator/lib/validateTileset.js
```diff
@@ -32,8 +40,67 @@ function validateNode(root, parent, resolve) {
...
+                            }
+                        } else if (magic === 'pnts') {
+                            var validPntsTest = validatePnts(tileBuffer);
+                            if (!validPntsTest.result) {
+                                resolve({
+                                    result: false,
+                                    message: 'invalid pnts'
+                                });
+                            }
+                        } else if (magic === 'cmpt') {
```
lilleyse: Just remove this for now, it can be added later.

validator/lib/validateTileset.js
```diff
@@ -32,8 +40,67 @@ function validateNode(root, parent, resolve) {
...
+                            message: 'Child occupies region greater than parent'
+                        }));
+                    }
+                }
+            }
+        }
+
+        if (defined(tile.content) && defined(tile.content.url)) {
+            tilePromises.push(readTile(tile.content.url)
+                .then(function(tileBuffer) {
```
lilleyse: Factor this out into a separate function called `validateContent`.

validator/lib/validateTileset.js
```diff
@@ -32,8 +40,67 @@ function validateNode(root, parent, resolve) {
         var tile = node.node;
         parent = node.parent;
 
+        if (defined(tile.content)) {
+            if (defined(tile.content.boundingVolume)) {
+                var region = tile.content.boundingVolume.region;
+                var parentRegion = tile.boundingVolume.region;
+                for (var i = 0; i < region.length; i++) {
+                    if (region[i] > parentRegion[i]) {
+                        tilePromises.push(resolve({
```
lilleyse: Remove `tilePromises.push` here.

validator/lib/validateTileset.js
```diff
@@ -3,6 +3,11 @@
 var Promise = require('bluebird');
 var Cesium = require('cesium');
 var defined = Cesium.defined;
+var readTile = require('../lib/readTile');
+var validateB3dm = require('../lib/validateB3dm');
+var validateI3dm = require('../lib/validateI3dm');
+var validatePnts = require('../lib/validatePnts');
+//var validateCmpt = require('../lib/validateCmpt');
```
lilleyse: Remove this line.

validator/lib/validateTileset.js
```diff
@@ -21,6 +26,9 @@ function validateTileset(tileset) {
 }
 
 function validateNode(root, parent, resolve) {
+
```
lilleyse: Remove empty line.

validator/lib/validateTileset.js
```diff
@@ -32,8 +40,67 @@ function validateNode(root, parent, resolve) {
...
+        }
+
+        if (defined(tile.content) && defined(tile.content.url)) {
+            tilePromises.push(readTile(tile.content.url)
+                .then(function(tileBuffer) {
+                    if (defined(tileBuffer)) {
+                        var magic = tileBuffer.toString('utf8', 0, 4);
+                        if (magic === 'b3dm') {
+                            var validateB3dmTest = validateB3dm(tileBuffer);
+                            if (!validB3dmTest.result) {
```
lilleyse: The naming here is wrong: validateB3dmTest vs validB3dmTest. Same with i3dm.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -77,3 +78,169 @@ describe('validateTileset', function() {
...
+        expect(validateBoundingVolume(invalidSphere)
+            .then(function(response) {
+                expect(response.result).toBe(false);
+                expect(response.message).toBe('Child bounding volume is not contained within parent');
+            }), done).toResolve();
+    });
+
+
+    it('reads a valid tileset', function(done) {
+        expect(validateTileset('./specs/data/TilesetGzipped/tileset2.json')
```
lilleyse: This should tileset.json right?

lilleyse:
I think the source of the problems is that the example tileset in `validateTilesetSpec` references non-existent content like b.b3dm, c.b3dm, etc. Eventually we'll need to check if a tileset references non-existent tiles, but for now I would just remove those contents.

sumitshyamsukha:
@lilleyse I got rid of the non-existent content, but I believe the issue now is that the file isn't being read into a tileset in `vaildateTileset`

lilleyse:
Ah ok. In the test before calling `validateTileset` you will need to call `readTileset`.

sumitshyamsukha:
@lilleyse Updated the test, but for the issue now is files such as `parent.b3dm` and `ll.b3dm` are in the directory `/specs/data/TilesetGzipped` along with `tileset.json` but `readTile` tries to look for them in the same directory. Should I also pass in the directory of the tileset as a param?

lilleyse:
> Should I also pass in the directory of the tileset as a param?

Yeah that sounds like a good idea.

sumitshyamsukha:
@lilleyse I added the path as an argument, and the test passes but I end up with the following error: 
`Unhandled rejection TypeError: Cannot read property 'addExpectationResult' of undefined
    at Env.fail (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:987:24)
    at Function.next.fail (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1903:19)
    at /Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/specs/matchers/expectPromise.js:14:26
    at tryCatcher (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/promise.js:504:31)
    at Promise._settlePromise (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/promise.js:561:18)
    at Promise._settlePromise0 (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/promise.js:606:10)
    at Promise._settlePromises (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/promise.js:681:18)
    at Async._drainQueue (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/async.js:138:16)
    at Async._drainQueues (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/async.js:148:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/sumitshyamsukha/Desktop/3d-tiles-tools/validator/node_modules/bluebird/js/release/async.js:17:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)`

pjcozzi:
@lilleyse can you help move this forward?

lilleyse:
Sorry for the delay @sumitshyamsukha.

In order to fix the error `validateTileset` will also need to accept that second parameter (maybe called `tilesetDirectory`, and then `tileContent.url` will need to become `path.join(tilesetDirectory, tileContent.url)`.

sumitshyamsukha:
@lilleyse The error seems to still come up even after changing that param. Could it be something with the tests?

lilleyse:
Yeah, the problem was with the construction of the test. The `, done).toResolve();` only needs to happen once. Also I needed to include `path` in `validateTileset`. Should be fixed now.

sumitshyamsukha:
Just pulled - seems to pass all tests now! Please let me know if there's anything else that needs to be addressed.

pjcozzi:
@lilleyse is this ready?

lilleyse:
Did some small cleanup - may require a bit more work but its good to merge into the `validator-master` branch.

Thanks @sumitshyamsukha!

Issue #32, opened 2016-11-18
Validator - validate batch table JSON against schema


lilleyse:
@JudyWeng 

From the [roadmap](https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/5): 

> Validate Batch Table is valid JSON and follows the schema (have a generic validation function for this).

For this you will need to extract the batch table JSON from the buffer, using the `batchTableJSONByteOffset` field of the header, and then use Node's Buffer class to convert to a string, and then `JSON.parse`. This whole step of getting the JSON from a buffer and byte offset can be made into its own helper file since it will be used by every tile format.

You should also extract the batch table binary. This doesn't need its own helper function because it should be a one-liner in Node.

Now add a new file called `validateBatchTable` which takes the batch table JSON and batch table binary. In here the first step will be to validate the JSON against the schema. I think the best idea for getting the schema is getting it directly from the github url where it is hosted. For this you can use Cesium's `loadJson` function. Once you have the schema, use a node library to validate the batch table JSON against the schema JSON. I believe a good library for this is https://www.npmjs.com/package/ajv.

The batch table schema is located here: https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/schema/batchTable.schema.json

PR #41, opened 2016-12-23
validating batch table

JudyWeng:
#32 
I think I have the helper methods to extract and validate the batch table, but the loadJson method from Cesium returns a promise and I'm still working on learning about promises and how to use it in the individual tile validation files.


pjcozzi:
@lilleyse will you be able to look at this soon?

lilleyse:
If I remember correctly, the main things to address here were:

* Move the batch table JSON and batch table binary extraction to the individual `validateB3dm`, `validateI3dm`, files. While there will be some duplicate code, I think it is worth it just so each tile type is self-contained.
* Avoid loading the batch table schema in `validateBatchTable`, since it would be much nicer to avoid promises entirely here. Rather it could be loaded right from the start in in `validateTileset` and eventually passed along to each tile in some sort of `schemas` object.

pjcozzi:
@JudyWeng any update here as well?

JudyWeng:
I had attempted the method of changing the tiles to promises first, which caused a lot of changes, but I'm going to try loading the batch table schema in validateTileset now instead to avoid having to change every validateB3dm, validateI3dm file.

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -1,9 +1,27 @@
...
 
+var loadJson = Cesium.loadJson;
+
 describe('validateB3dm', function() {
+    var batchSchema;
+    beforeAll(function(done) {
+        console.log("beforeAll");
 
+        loadJson('https://raw.githubusercontent.com/AnalyticalGraphicsInc/3d-tiles/master/schema/batchTable.schema.json')
+            .then(function(schema) {
```
JudyWeng: I tried to change the validate files to take the batch table schema in as a parameter to avoid having to change everything  into an asynchronous process, as suggested previously. However, it seems like loadJson() is returning the "XMLHttpRequest i undefined" error. I have tried both the url above and this url: "https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/schema/batchTable.schema.json" to load the Json. Am I not using this function properly?

lilleyse: Hm, `XMLHttpRequest` must only be supported in browsers and not Node. You could try looking at [http.request](https://nodejs.org/api/http.html#http_http_request_options_callback).

JudyWeng: If I choose to use http.request it would require me to change loadJson() though wouldn't it?

pjcozzi: @lilleyse any thoughts here?

lilleyse: We discussed this offline, that using http.request (or alternatives) is fine to replace loadJson.

JudyWeng:
I wrote the test cases for the binary table validation for b3dm, i3dm, and pnts, but not cmpt yet because that validation branch has yet to be merged into validator-master. Please let me know if there are more edge cases I have yet to cover.

lilleyse:
I'll look at both later today.

validator/lib/validateB3dm.js
```diff
@@ -19,7 +21,14 @@ function validateB3dm(content) {
     }
 
     if (!Buffer.isBuffer(content)) {
-        throw new DeveloperError('content must be of type buffer');
+        throw new DeveloperError('b3dm content must be of type buffer');
+    }
+
+    if(content.length < 2) {
```
lilleyse: Should this be 24?

lilleyse: Style: `if(` -> `if (`.

Fix throughout.

validator/lib/validateB3dm.js
```diff
@@ -29,27 +38,90 @@ function validateB3dm(content) {
...
+
+    if(message === '') {
+        message = 'successfully extracted batch table'
+    }
+
+    return {
+        batchTableJSON: batchTableJSON,
+        batchTableBinary: batchTableBinary,
+        message: message
+    };
```
lilleyse: This section may be able to be condensed by extracting the sections, and then verifying they are correct.

For example - https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/tools/lib/extractB3dm.js#L44-L46. You could do something like that and then verify the buffers are the correct size. Ideally the batch table extraction/validation could even go inline with the rest of the b3dm checking.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+var Cesium = require('cesium');
+var Ajv = require('ajv');
```
lilleyse: Add requires in alphabetical order.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,29 @@
...
+
+module.exports = validateBatchTable;
+
+/**
+ * Checks if provided buffers follow the batch table schema
+ *
+ * @param {Object} schema - A JSON object containing the schema for the batch table.
+ * @param {Buffer} batchTableJSON - A buffer containing the JSON Header of a batch table
```
lilleyse: `JSON Header of a batch table` -> `batch table JSON`. It's not a buffer, right?

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Buffer} batchTableBinary - A buffer containing the binary body of a batch table
```
lilleyse: `binary body of a batch table` -> `batch table binary` is simpler.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,29 @@
...
+ * @returns {Object} An object with two parameters - (1) a boolean for whether the batch table follows the schema
+ *                                                   (2) a message to indicate the validation result
+ */
+function validateBatchTable(schema, batchTableJSON, batchTableBinary) {
+    var ajv = new Ajv();
+    var validation = ajv.validate(schema, batchTableJSON);
```
lilleyse: `valid` instead of `validation`?

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -1,23 +1,43 @@
...
+    it('returns false if the b3dm has wrong byteLength', function() {
         expect(validateB3dm(createWrongByteLength()).result).toBe(false);
     });
 
-    it('returns true if b3dm tile matches spec', function() {
+    it('validates b3dm tile matches spec', function() {
         expect(validateB3dm(createB3dmTile()).result).toBe(true);
     });
+
+    it('validates b3dm tile with batch table JSON header matches spec', function() {
```
lilleyse: The wording for these tests is awkward.

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -33,37 +53,140 @@ function createB3dmTile() {
...
-    var header = new Buffer(24);
+    var header = createB3dmTile();
     header.write('xxxx', 0); // magic
-    header.writeUInt32LE(1, 4); // version
-    header.writeUInt32LE(header.length, 8); // byteLength
-    header.writeUInt32LE(0, 12); // batchTableJSONByteLength
-    header.writeUInt32LE(0, 16); // batchTableBinaryByteLength
-    header.writeUInt32LE(0, 20); // batchLength
-
+    console.log('checking:\n' + header + '\n');
```
lilleyse: Remove this.

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -33,37 +53,140 @@ function createB3dmTile() {
...
     return header;
+}
+
+function createB3dmBatchJson() {
+    var header = createB3dmTile();
+    var batchJSON = createValidBatchTableJSON();
+    header.writeUInt32LE(header.length + batchJSON.length, 8); // byteLength
+    header.writeUInt32LE(batchJSON.length, 12); // batchTableJSONByteLength
+
+    return Buffer.concat([header, batchJSON]);
```
lilleyse: Better to avoid condensing the name, `batchTableJSON` while longer is easier to follow.

lilleyse: The name of the function could also be `createB3dmWithBatchTableJson`.

lilleyse: Also make sure to be consistent with `Json` vs `JSON`.

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -33,37 +53,140 @@ function createB3dmTile() {
...
+
+    return Buffer.concat([header, batchTable.buffer]);
+}
+
+function createValidBatchTableJSON() {
+    var batchJson = {
+        "id":[0,1,2],
+        "longitude":[-1.3196595204101946,-1.3196567190670823,-1.3196687138763508],
+        "height":[8,14,14]
+    };
```
lilleyse: No need to wrap each key in quotes.

validator/specs/lib/validateB3dmSpec.js
```diff
@@ -33,37 +53,140 @@ function createB3dmTile() {
...
+}
+
+function createValidBatchTableBinary() {
+    var batchJson = {
+        "id" : [0, 1, 2],
+        "longitude" :[-1.3196595204101946,-1.3196567190670823,-1.3196687138763508],
+        "height" : {
+            "byteOffset" : 12,
+            "componentType" : "UNSIGNED_INT",
+            "type" : "SCALAR"
```
lilleyse: Use single quotes for `UNSIGNED_INT` and `SCALAR`.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+    'VEC2': 2,
+    'VEC3': 3,
+    'VEC4': 4
+};
+
+/**
+ * Checks if provided buffers follow the batch table schema
+ *
+ * @param {Object} schema - A JSON object containing the schema for the batch table.
+ * @param {JSON} batchTableJSON - Batch table JSON
```
lilleyse: I don't think this is a valid jsdoc type, use `Object` instead.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+            var binaryBodyLength = batchTableBinary.length;
+            var batchLength = batchTableJSON["id"].length;
+            var totalOffset = 0;
+            for (var key in batchTableJSON) {
+                if (!batchTableJSON.hasOwnProperty(key)) {
+                    continue;
+                }
+
+                var val = batchTableJSON[key];
+                if (Object.prototype.toString.call(val) === '[object Array]') {
```
lilleyse: Also validate that the array length equals batch length.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+        if(defined(batchTableBinary) && (batchTableBinary.length > 0)) {
+            var binaryBodyLength = batchTableBinary.length;
+            var batchLength = batchTableJSON["id"].length;
+            var totalOffset = 0;
+            for (var key in batchTableJSON) {
+                if (!batchTableJSON.hasOwnProperty(key)) {
+                    continue;
+                }
+
+                var val = batchTableJSON[key];
```
lilleyse: I feel like `property` is a more fitting name.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+function validateBatchTable(schema, batchTableJSON, batchTableBinary) {
+    var ajv = new Ajv();
+    var validSchema = ajv.validate(schema, batchTableJSON);
+    var valid = validSchema;
+    var message;
+
+    if(validSchema) {
+        if(defined(batchTableBinary) && (batchTableBinary.length > 0)) {
+            var binaryBodyLength = batchTableBinary.length;
+            var batchLength = batchTableJSON["id"].length;
```
lilleyse: The property `id` is not guaranteed to exist. The batch length is now in the feature table: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel#global-semantics

validator/lib/validateI3dm.js
```diff
@@ -19,7 +21,14 @@ function validateI3dm(content) {
...
 
     if (!Buffer.isBuffer(content)) {
-        throw new DeveloperError('content must be of type buffer');
+        throw new DeveloperError('i3dm content must be of type buffer');
+    }
+
+    if (content.length < 32) {
+        return {
+            result : false,
+            message: 'i3dm tile header must be 28 bytes'
```
lilleyse: 32

validator/lib/validateI3dm.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 var Cesium = require('cesium');
+var batchTableSchema = require('../specs/data/schema/batchTable.schema.json');
```
lilleyse: Overall same comments here as `validateB3dm`.

validator/lib/validateI3dm.js
```diff
@@ -30,31 +39,60 @@ function validateI3dm(content) {
...
             result : false,
-            message: 'Header has an invalid gltfFormat field. Expected gltfFormat = 0 or 1. Found gltfFormat = ' + gltfFormat
+            message: 'i3dm tile has an invalid gltfFormat field. Expected gltfFormat = 0 or 1. Found gltfFormat = ' + gltfFormat
         };
     }
 
+    var offset = 32;
+    var featureTableJSONByteLength = content.readUInt32LE(12);
+    var featureTableBinaryByteLength = content.readUInt32LE(16);
+    offset += featureTableJSONByteLength + featureTableBinaryByteLength;
```
lilleyse: Style: move this a few lines below so it doesn't cut off the header value reads.

validator/specs/lib/validateBatchTableSpec.js
```diff
@@ -0,0 +1,106 @@
...
+
+    return batchTableJSON;
+}
+
+function createInvalidBatchTableJSON() {
+    var batchTableJSON = {
+        id:[0],
+        longitude:[-1.3196595204101946],
+        height:8
+    };
```
lilleyse: For any of these, add some spacing 

`id : [0],`

validator/specs/lib/validateBatchTableSpec.js
```diff
@@ -0,0 +1,106 @@
...
+        height : {
+            "byteOffset" : 0,
+            "componentType" : 'UNSIGNED_INT',
+            "type" : 'SCALAR'
+        },
+        width : {
+            "byteOffset" : 12,
+            "componentType" : 'UNSIGNED_INT',
+            "type" : 'SCALAR'
+        }
```
lilleyse: Fine to do:
```
width : {
    byteOffset : 12,
    componentType : 'UNSIGNED_INT',
    type : 'SCALAR'
}
```

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+                var numberOfComponents = val.type;
+
+                if (byteOffset < totalOffset) {
+                    valid = false;
+                    message = 'batch table property ' + key + ' has offset within another property\'s range at ' + byteOffset;
+                    break;
+                }
+
+                var propertyByteLength = batchLength * componentTypeSize[componentType] * typeSize[numberOfComponents];
+                if (propertyByteLength + byteOffset > binaryBodyLength) {
```
lilleyse: Also worth validating that `componentType` and `type` are actually correct.

validator/lib/validateBatchTable.js
```diff
@@ -0,0 +1,89 @@
...
+    'VEC3': 3,
+    'VEC4': 4
+};
+
+/**
+ * Checks if provided buffers follow the batch table schema
+ *
+ * @param {Object} schema - A JSON object containing the schema for the batch table.
+ * @param {JSON} batchTableJSON - Batch table JSON
+ * @param {Buffer} batchTableBinary - A buffer containing the batch table binary
```
lilleyse: Since `batchTableBinary` is optional, change to `[batchTableBinary]`

validator/specs/lib/validateBatchTableSpec.js
```diff
@@ -0,0 +1,106 @@
...
+        expect(validateBatchTable(batchTableSchema, batchTableJSON).validation).toBe(false);
+    });
+
+    it('returns false if batch table JSON does not match schema with binary body', function() {
+        var batchTableJSON = createInvalidBatchTableJSONBinary();
+        var batchTableBinary = createValidBatchTableBinary();
+        expect(validateBatchTable(batchTableSchema, batchTableJSON, batchTableBinary).validation).toBe(false);
+    });
+
+    it('byteoffset out of bounds', function() {
```
lilleyse: Make these two slightly more descriptive - `Binary property byteOffset out of bounds`

validator/specs/lib/validateCmptSpec.js
```diff
@@ -186,3 +199,89 @@ function createCmptInvalidCombination() {
...
+function createCmptWithBatchTable() {
+    //cmpt1[b3dm, cmpt2[cmpt3[pnts], i3dm]]
+    var innerB3dmTile = createB3dmTile();
+    var innerPntsTile = createPntsTile();
+    var innerI3dmTile = createI3dmTile();
+
+    var batchTable = createBatchTableBinary();
+    innerB3dmTile.writeUInt32LE(innerB3dmTile.length + batchTable.buffer.length, 8); // byteLength
+    innerB3dmTile.writeUInt32LE(batchTable.batchTableJSONByteLength, 12); // batchTableJSONByteLength
+    innerB3dmTile.writeUInt32LE(batchTable.batchTableBinaryByteLength, 16);
```
lilleyse: 12 and 16 are now offsets for the feature table

JudyWeng:
I've updated the old tests to the new b3dm specs and they pass when I run them. I'm not sure what else might be causing the checks to fail...

lilleyse:
You can ignore that, Travis is broken for this repo right now due to the directory structure.

lilleyse:
I'm late here, but thanks for all the work @JudyWeng!

Issue #39, opened 2016-12-13
Extract tiles from Composite tiles


pjcozzi:
Add a stage to recursively extract tiles from composite tiles, including support for composites of composites.

pjcozzi:
Functions added in #42.  CLI extracts glbs.

PR #42, opened 2017-01-03
Added tool to extract glbs from a cmpt

lilleyse:
I forgot I had this lying around - should be ready to review.


tools/bin/3d-tiles-tools.js
```diff
@@ -265,6 +265,53 @@ function readB3dmWriteGlb(inputPath, outputPath, force) {
...
+
+function readCmptWriteGlb(inputPath, outputPath, force) {
+    outputPath = defaultValue(outputPath, inputPath.slice(0, inputPath.length - 5));
+    return fsReadFile(inputPath)
+        .then(function(data) {
+            var tiles = extractCmpt(data);
+            var tilesLength = tiles.length;
+            var tilePaths = new Array(tilesLength);
+            if (tilesLength === 0) {
+                console.log('No glbs found in this cmpt tile.');
```
pjcozzi: "this?"

Why not output the path?

pjcozzi:
@lilleyse when I first saw this, I thought we should change this to be a stage that extracts tiles from composites, that is then piped into another stage that extracts glbs from tiles.

What do you think?

What you have here could be fine for today's needs (debugging glbs) and we could just close #39?

pjcozzi:
Code and tests are good.

lilleyse:
> @lilleyse when I first saw this, I thought we should change this to be a stage that extracts tiles from composites, that is then piped into another stage that extracts glbs from tiles.

I think that's a good idea and better separation of tasks. I made the changes, but instead of making a `tilesToGlb` stage I just added a helper function in `3d-tiles-tools.js`.

Issue #36, opened 2016-12-05
Optimize glTF models in i3dm tiles using gltf-pipeline


pjcozzi:
Similar to #34.

PR #46, opened 2017-01-19
Optimize i3dm

shehzan10:
Added:
* Function to convert glb to i3dm
* Function to read and call optimization on i3dm
* Tests


lilleyse:
Fixes #36

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Generates a new Buffer representing a i3dm asset.
+ *
+ * @param {Buffer} glbBuffer A buffer containing a binary glTF asset.
+ * @param {Buffer} [featureTableJSONBuffer] A buffer containing the batch table to use for the b3dm asset.
+ * @param {Buffer} [featureTableBinaryBuffer] A buffer containing the accompanying binary batch table section for the b3dm asset.
+ * @param {Buffer} [batchTableJSONBuffer] A buffer containing the batch table to use for the b3dm asset.
+ * @param {Buffer} [batchTableBinaryBuffer] A buffer containing the accompanying binary batch table section for the b3dm asset.
+ * @param {Number} [gltfFormat] The gltf format for the body of the glTF field.
```
lilleyse: `gltfFormat` doesn't need to be passed in since this tool always embeds the gltf. I think it's fine to keep the changes in `extractI3dm` though.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+ * @returns {Buffer} Buffer representing the b3dm asset.
+ */
```
lilleyse: Some of the doc here still has the b3dm terminology.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+function glbToI3dm(glbBuffer, featureTableJSONBuffer, featureTableBinaryBuffer, batchTableJSONBuffer, batchTableBinaryBuffer, gltfFormat) {
+    if (!defined(glbBuffer)) {
+        throw new DeveloperError('glbBuffer is not defined.');
+    }
+
+    batchTableJSONBuffer = defaultValue(batchTableJSONBuffer, new Buffer(0));
```
lilleyse: For this and anywhere else in the codebase that create `new Buffer` can you change to `Buffer.alloc`? The direct constructors are deprecated.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+    batchTableBinaryBuffer = defaultValue(batchTableBinaryBuffer, new Buffer(0));
+    featureTableJSONBuffer = defaultValue(featureTableJSONBuffer, new Buffer(0));
```
lilleyse: For `i3dm` the feature table is required since both `POSITION` and `INSTANCES_LENGTH` are required. It would be fine to add a single instance whose position is [0,0,0]. https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Instanced3DModel

Perhaps that should be done in `3d-tiles-tools.js` before getting here.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,70 @@
...
+    var glbBuffer;
+    beforeAll(function(done) {
+        fsReadFile(glbPath)
+            .then(function(data) {
+                glbBuffer = data;
+                done();
+            });
+    });
+
+    it('generates a basic I3dm header for a given buffer representing a glb', function() {
```
lilleyse: I3dm to i3dm.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+    featureTableBinaryBuffer = defaultValue(featureTableBinaryBuffer, new Buffer(0));
+    gltfFormat = defaultValue(gltfFormat, 1);
+
+    var byteLength = 32 + glbBuffer.length + featureTableJSONBuffer.length + featureTableBinaryBuffer.length + batchTableJSONBuffer.length + batchTableBinaryBuffer.length;
+
+    var header = new Buffer(32);
+    header.write('i3dm', 0); // magic
+    header.writeUInt32LE(1, 4); // version
+    header.writeUInt32LE(byteLength, 8); // byteLength - length of entire tile, including header, in bytes
+    header.writeUInt32LE(featureTableJSONBuffer.length, 12); // featureTableJSONByteLength - length of feature table JSON section in bytes. (0 for basic, no batches)
```
lilleyse: Remove the `(0 for basic, no batches)` comment for all these, it doesn't really apply for i3dm as much.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,45 @@
...
+    header.writeUInt32LE(featureTableBinaryBuffer.length, 16); // featureTableBinaryByteLength - length of feature table binary section in bytes. (0 for basic, no batches)
+    header.writeUInt32LE(batchTableJSONBuffer.length, 20); // batchTableJSONByteLength - length of feature table JSON section in bytes. (0 for basic, no batches)
+    header.writeUInt32LE(batchTableBinaryBuffer.length, 24); // batchTableBinaryByteLength - length of feature table binary section in bytes. (0 for basic, no batches)
```
lilleyse: For these two `feature table` should be `batch table`.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,70 @@
...
+        expect(header.readUInt32LE(28)).toEqual(1); // gltfFormat - format of the body field
+    });
+
+    it('throws an error if no glbBuffer is provided', function() {
+        expect(function() {
+            glbToI3dm();
+        }).toThrowError('glbBuffer is not defined.');
+    });
+
+    it('convert I3dm to glb and back to I3dm', function() {
```
lilleyse: I3dm to i3dm here too.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,70 @@
...
+        fsReadFile(i3dmPath)
+            .then(function(fileBuffer) {
+                var i3dm = extractI3dm(fileBuffer);
+                var i3dmOut = glbToI3dm(i3dm.glb, i3dm.featureTable.json, i3dm.featureTable.binary,
+                                        i3dm.batchTable.json, i3dm.batchTable.binary, i3dm.header.gltfFormat);
+                expect(i3dm).toEqual(i3dmOut);
+            });
+    });
```
lilleyse: Since this test includes a promise, it needs to check that it is resolved. For example like `optimizeGlbSpec`:

```javascript
    it('optimizes a glb using the gltf-pipeline', function(done) {
        expect(optimizeGlb(buffer)
            .then(function(optimizedGlb) {
                expect(optimizedGlb).not.toEqual(buffer);
            }), done).toResolve();
    });
```

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,70 @@
...
+
+    it('throws an error if no glbBuffer is provided', function() {
+        expect(function() {
+            glbToI3dm();
+        }).toThrowError('glbBuffer is not defined.');
+    });
+});
```
lilleyse: Duplicate of a test above.

tools/bin/3d-tiles-tools.js
```diff
@@ -88,12 +90,7 @@ var argv = yargs
             description: 'All arguments after this flag will be passed to gltf-pipeline as command line options.'
         }
     })
-    .command('gzip', 'Gzips the input tileset directory.', {
-        't': {
-            alias: 'tilesOnly',
-            description: 'Only tile files (.b3dm, .i3dm, .pnts, .vctr) should be gzipped.'
-        }
-    })
```
shehzan10: Oops. Not sure how that slipped through.

tools/lib/glbToI3dm.js
shehzan10: I wanted to keep it around in case we decide to support both. Should I still remove it?

tools/lib/glbToI3dm.js
shehzan10: I'll add this requirement. However, I think this should stay here as the last line of defense. Otherwise, we should throw an error if the feature table is undefined.

What would be the better scenario?

tools/lib/glbToI3dm.js
lilleyse: Yeah it's fine to remove for now.

tools/lib/glbToI3dm.js
lilleyse: The main reason I think its better outside of here is because even though it is required for the i3dm to be valid, it is nice to keep this file really direct and not make assumptions about the data sent to it.

shehzan10: By extension, we are throwing an error for [`gltfByteLength === 0`](https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/tools/lib/extractI3dm.js#L50-L53).

So unless `glbToI3dm` is directly called by the user, bypassing `3d-tiles-tools.js`, the error would be thrown before glbToI3dm is called by `3d-tiles-tools.js`. Thus it would never actually execute the fallback to set  `POSITION` and `[0, 0, 0]` as you suggested.

shehzan10:
@lilleyse resolved.

tools/bin/3d-tiles-tools.js
```diff
@@ -250,6 +257,29 @@ function readGlbWriteB3dm(inputPath, outputPath, force) {
...
+                console.log('File ' + outputPath + ' already exists. Specify -f or --force to overwrite existing file.');
+                return;
+            }
+            return fsReadFile(inputPath)
+                .then(function(data) {
+                    // Set I3dm spec requirements
+                    var featureTable = { INSTANCES_LENGTH : 0,
+                                         EAST_NORTH_UP : true,
+                                         POSITION : { byteOffset : 0 }
+                                       };
```
lilleyse: Usually we format JSON in-line with the rest of the code, like:

```javascript
var featureTable = {
    INSTANCES_LENGTH : 0,
    EAST_NORTH_UP : true,
    POSITION : {
        byteOffset : 0
    }
};
```
Good call on the EAST_NORTH_UP.

tools/bin/3d-tiles-tools.js
```diff
@@ -250,6 +257,29 @@ function readGlbWriteB3dm(inputPath, outputPath, force) {
...
+                    var featureTableJSONBuffer = Buffer.alloc(JSON.stringify(featureTable));
```
lilleyse: Ah.. so the one exception is to use `Buffer.from` for strings.

tools/bin/3d-tiles-tools.js
```diff
@@ -250,6 +257,29 @@ function readGlbWriteB3dm(inputPath, outputPath, force) {
...
+                    var featureTableBinaryBuffer = Buffer.alloc(3, 0); // [0, 0, 0]
```
lilleyse: Should be 12 bytes - positions are floats.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,86 @@
...
+        fsReadFile(glbPath)
+            .then(function(data) {
+                glbBuffer = data;
+                var featureTable = { INSTANCES_LENGTH : 0,
+                                     EAST_NORTH_UP : true,
+                                     POSITION : { byteOffset : 0 }
+                                   };
+                var featureTableString = JSON.stringify(featureTable);
+                featureTableJSONBuffer = Buffer.alloc(featureTableString.length, featureTableString);
+                featureTableBinaryBuffer = Buffer.alloc(3, 0); // [0, 0, 0]
```
lilleyse: Similar comments as above here.

tools/lib/glbToI3dm.js
```diff
@@ -0,0 +1,50 @@
...
+        throw new DeveloperError('glbBuffer is not defined.');
+    }
+
+    if (!defined(featureTableJSONBuffer)) {
+        throw new DeveloperError('featureTableJSONBuffer is not defined.');
+    }
+
+    if (!defined(featureTableBinaryBuffer)) {
+        throw new DeveloperError('featureTableBinaryBuffer is not defined.');
+    }
```
lilleyse: I'm fine with removing these two defined checks. I wouldn't consider it an error if the feature table is missing - it will produce invalid i3dms, but that should be up to the code that calls this.

tools/bin/3d-tiles-tools.js
lilleyse: Sorry... I take that back with EAST_NORTH_UP. That's a good value to set if the instance positions are located on the globe to orient them property, but for this the position is at zero and so EAST_NORTH_UP shouldn't be set.

lilleyse:
Why this is ready it would be good to test that the results are working correctly in Cesium.

For `glbToI3dm`, run a random glb through the stage. On the 3d-tiles branch in Cesium go to the `Specs/Data/Cesium3DTiles/InstancedWithTransform/` and replace the i3dm there with the newly generated one. The reason for using the transform tileset is because the instance position is set to 0,0,0 and the transform in tileset.json will transform it to the globe's surface.

For `optimizeI3dm`, test with `instancedTile.i3dm` in `specs/data`. Then replace the i3dm inside `InstancedWithBatchTable`.

To see the results, open the `3D Tiles` sandcastle and edit the urls to point to these tilesets.


When this is ready, run an i3dm through optimizeI3dm and run a glb through glbToI3dm.



On the 3d-tiles branch in Cesium go to the `Specs/Data/Cesium3DTiles/InstancedWithBatchTable/` and replace the i3dm there with the newly generated one. 

one nice way to test that everything is working is to bring the  try using the `glbToI3dm`

tools/lib/glbToI3dm.js
shehzan10: It's possible that `featureTableJSONBuffer` and `featureTableBinaryBuffer` are `undefined` are of a type that does not have the `length` property defined. In that case an exception would be produced. Would this be an acceptable situation not to handle?

lilleyse:
Just to summarize from some offline discussion:

Let's replace `instanced.i3dm` with a better file. Using the instanced trees is just asking for issues. Replacing it with Cesium's `Specs/Data/Cesium3DTiles/InstancedWithBatchTable/instancedWithBatchTable.i3dm` should be good.

Cesium has trouble loading the the model that goes through `glbToI3dm`. I haven't looked too hard but the error is due to improper byte alignment of the typed array, reference: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/FeatureTable#implementation-notes

lilleyse:
Make sure to handle this comment as well in case you missed it: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/46#discussion_r97138689

pjcozzi:
Also add instructions for this to the README.md.

shehzan10:
> @shehzan10 adding a breakpoint to Instanced3DModel3DTileContent.initialize should help track it down.

So here are the steps I did:
1. I converted `CesiumTexturedBox.glb` into an i3dm, and loaded into Cesium - nothing shows up.
2. I then converted the `InstancedWithBatchTable.i3dm` file from Cesium into a `glb` then back into a `i3dm` - and nothing shows up.
3. I added the padding code from [generator PR](https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/43/files) into the tools lib. And then repeated steps 1 and 2. This time I got errors with the following stack trace:
```
DeveloperError: Operator ">=" requires number arguments. Arguments are undefined and 83.
Error
    at new DeveloperError (http://localhost:8080/Source/Core/DeveloperError.js:44:19)
    at Node._evaluateGreaterThanOrEquals (http://localhost:8080/Source/Scene/Expression.js:1098:19)
    at Expression.evaluate (http://localhost:8080/Source/Scene/Expression.js:208:39)
    at ConditionsExpression.evaluateColor (http://localhost:8080/Source/Scene/ConditionsExpression.js:144:41)
    at applyStyleWithBatchTable (http://localhost:8080/Source/Scene/Cesium3DTileStyleEngine.js:122:41)
    at styleContent (http://localhost:8080/Source/Scene/Cesium3DTileStyleEngine.js:104:13)
    at styleCompositeContent (http://localhost:8080/Source/Scene/Cesium3DTileStyleEngine.js:94:13)
    at Cesium3DTileStyleEngine.applyStyle (http://localhost:8080/Source/Scene/Cesium3DTileStyleEngine.js:77:21)
    at updateTiles (http://localhost:8080/Source/Scene/Cesium3DTileset.js:1613:30)
    at Cesium3DTileset.update (http://localhost:8080/Source/Scene/Cesium3DTileset.js:1769:9)
```
> Note: I did update the `tileset.json` files in cesium.

I debugged and checked that the byte offsets are being loaded with the padding when I used it and all other things look good in the `initialize` function too.

lilleyse:
Ah, `DeveloperError: Operator ">=" requires number arguments. Arguments are undefined and 83.` is an error related to styling. Remove the style from the demo and it should work. It's referencing a non-existent property.

shehzan10:
So removing the style seems to fix that error. However, even the padded model files don't seem to be showing up.

I also tried with a padding of 4 bytes (previously 8 bytes). But still nothing. With 4 byte padding, `[0, 0, 0]` would not need to be padded, whereas with 8 bytes, it would get padded to 16 bytes, hence adding another 0. So I tried 4 bytes.

lilleyse:
Yes 4 byte padding should be sufficient. Where does the load fail in Cesium?

shehzan10:
@lilleyse do you want to move the loading discussion into a separate issue? I think since this PR is not handling that part of the code, we can debug and open a new PR for that.

lilleyse:
We should keep this open until we've ruled out the issue for sure.

shehzan10:
Summary of offline discussion: The model being used was the flat tree model and almost impossible to see (possible due to a copying error). I have removed that model and replaced it with the `instancedWithTransform.i3dm` model from Cesium 3d-tiles branch.
We also looking into the `tileset.json` files in cesium not having a transform.

I have fixed the issues and models and tested with model inspector and cesium. This PR should be good to go.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,76 @@
...
+                glbBuffer = data;
+                var featureTable = {
+                    INSTANCES_LENGTH : 1,
+                    POSITION : {
+                        byteOffset : 0
+                    }
+                };
+                var featureTableString = JSON.stringify(featureTable);
+                featureTableJSONBuffer = Buffer.from(featureTableString);
+                featureTableBinaryBuffer = Buffer.alloc(12, 0); // [0, 0, 0]
```
lilleyse: It would be best if the specs also used the `getBufferPadded` and `getJsonBufferPadded` functions.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,76 @@
...
+    it('convert i3dm to glb and back to i3dm', function(done) {
+        expect(fsReadFile(i3dmPath)
+                .then(function(fileBuffer) {
+                    var i3dm = extractI3dm(fileBuffer);
+                    var i3dmOut = glbToI3dm(i3dm.glb, i3dm.featureTable.json, i3dm.featureTable.binary,
+                                            i3dm.batchTable.json, i3dm.batchTable.binary);
+                    expect(i3dm).toEqual(extractI3dm(i3dmOut));
+                    expect(i3dmOut).toEqual(fileBuffer);
+                }), done).toResolve();
+    });
```
lilleyse: Too much indentation in this test.

tools/specs/lib/glbToI3dmSpec.js
```diff
@@ -0,0 +1,76 @@
...
+            glbToI3dm();
+        }).toThrowError('glbBuffer is not defined.');
+    });
+
+    it('convert i3dm to glb and back to i3dm', function(done) {
+        expect(fsReadFile(i3dmPath)
+                .then(function(fileBuffer) {
+                    var i3dm = extractI3dm(fileBuffer);
+                    var i3dmOut = glbToI3dm(i3dm.glb, i3dm.featureTable.json, i3dm.featureTable.binary,
+                                            i3dm.batchTable.json, i3dm.batchTable.binary);
```
lilleyse: Looks cleaner as one line, even though it exceeds the usual char limit.

tools/specs/lib/extractI3dmSpec.js
```diff
@@ -5,7 +5,7 @@ var extractI3dm = require('../../lib/extractI3dm');
 
 var fsReadFile = Promise.promisify(fs.readFile);
 
-var i3dmPath = './specs/data/instancedTile.i3dm';
+var i3dmPath = './specs/data/instancedWithTransform.i3dm';
```
lilleyse: Instead use the i3dm in `Specs/Data/Cesium3DTiles/Instanced/InstancedWithBatchTable`, that way it won't be confusing if someone tries to load this i3dm without specifying a transform. You can still use the name `instancedTile.i3dm` though.

lilleyse:
I confirmed that both modes are working in Cesium.

shehzan10:
Fixed the suggested changes.

lilleyse:
Awesome! Looks great now.

PR #53, opened 2017-02-18
fix typo

foobarbecue:


lilleyse:
Thanks for the fix.

I know this is a small PR, but before merging we require a CLA: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

PR #55, opened 2017-02-20
Make outputFile optional for tileset2sqlite

lilleyse:
To be consistent with the other tools the output file is now optional for tileset2sqlite.

This also fixes an error if the .3dtiles file already exists and -f is set, in which case it opens the existing database and tries to write to it. I just delete the file if it exists.

None of the code after `// Create the database.` has changed, just indented.


pjcozzi:
Looks good!

PR #57, opened 2017-02-22
Clean up 3d-tiles-tools.js

lilleyse:
Just some cleanup in the 3d-tiles-tools.js file. The one breaking change is removing the `-z` option for `optimizeB3dm` to gzip the b3dm. Instead it will gzip automatically if the input file is gzipped.

### DIFF
--- a/tools/bin/3d-tiles-tools.js
+++ b/tools/bin/3d-tiles-tools.js
@@ -1,12 +1,13 @@
 #!/usr/bin/env node
 'use strict';
 var Cesium = require('cesium');
-var GltfPipeline = require('gltf-pipeline');
-var Promise = require('bluebird');
 var fsExtra = require('fs-extra');
+var GltfPipeline = require('gltf-pipeline');
 var path = require('path');
+var Promise = require('bluebird');
 var yargs = require('yargs');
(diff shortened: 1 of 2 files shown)


shehzan10:
Do you think it would make it useful to make `force` a part of `options` across all the function and make options as standard instead?

shehzan10:
Closed by mistake.

lilleyse:
> Do you think it would make it useful to make force a part of options across all the function and make options as standard instead?

I think since this is all in the bin script its ok for now. If we find ourselves adding more of the same options to every function we could go with that approach.

shehzan10:
Just to confirm, if the input B3dm is unzipped and the user wanted it to be zipped, then the user would need to call both `optimizeB3dm` followed by `gzip`?

The code changes look solid.

lilleyse:
> Just to confirm, if the input B3dm is unzipped and the user wanted it to be zipped, then the user would need to call both optimizeB3dm followed by gzip?

We don't actually have a command that gzips individual tiles, just tilesets. My main reason for removing the option is we don't give that option for any of the other tools, like `glbToB3dm` and others.

PR #58, opened 2017-02-24
Add tools to index.js

shehzan10:


tools/index.js
```diff
@@ -1,6 +1,11 @@
 module.exports = {
+    extractB3dm : require('./lib/extractB3dm'),
+    extractCmpt : require('./lib/extractCmpt'),
+    extractI3dm : require('./lib/extractI3dm'),
     glbToB3dm : require('./lib/glbToB3dm'),
     glbToI3dm : require('./lib/glbToI3dm'),
     gzipTileset : require('./lib/gzipTileset'),
-    pipeline : require('./lib/runPipeline')
+    optimizeGlb : require('./lib/gzipTileset'),
```
lilleyse: I don't see the need to expose this one as it is equivalent to just including `gltf-pipeline` as a library.

shehzan10: Fixed. Also caught a copy-paste mistake in that. So thanks.

shehzan10: Oops. Sorry changed the wrong thing.

tools/index.js
```diff
@@ -1,6 +1,10 @@
 module.exports = {
+    extractB3dm : require('./lib/extractB3dm'),
+    extractCmpt : require('./lib/extractCmpt'),
+    extractI3dm : require('./lib/extractI3dm'),
     glbToB3dm : require('./lib/glbToB3dm'),
     glbToI3dm : require('./lib/glbToI3dm'),
-    gzipTileset : require('./lib/gzipTileset'),
-    pipeline : require('./lib/runPipeline')
+    optimizeGlb : require('./lib/optimizeGlb'),
+    pipeline : require('./lib/runPipeline'),
```
lilleyse: Better to match the naming. `pipeline` -> `runPipeline`.

lilleyse:
`gzipTileset` shouldn't be removed from here.

PR #62, opened 2017-03-08
.3dtiles to tileset command

### LINKED ISSUE #56: Unpack a .3dtiles database file
It would be convenient to have a stage that takes a `.3dtiles` file and unpacks it to a tileset folder. It should ungzip tiles in the database.

lilleyse:
Fixes #56

Breaking changes: `tileset2sqlite3` is now called `tilesetToDatabase`. This will need to be fixed elsewhere once a new version of 3d-tiles-tools is published.

@tfili can you quickly check that the sql commands are sane?

@shehzan10 can you do a full review? No rush on this.


tools/lib/databaseToTileset.js
```diff
@@ -0,0 +1,56 @@
...
+
+var fsExtraOutputFile = Promise.promisify(fsExtra.outputFile);
+
+var defaultValue = Cesium.defaultValue;
+var defined = Cesium.defined;
+var DeveloperError = Cesium.DeveloperError;
+
+module.exports = tilesetToDatabase;
+
+function tilesetToDatabase(inputFile, outputDirectory) {
```
tfili: Probably should be `databaseToTileset`.

tools/lib/databaseToTileset.js
```diff
@@ -0,0 +1,56 @@
...
+            }
+            return Promise.all(promises);
+        })
+        .finally(function() {
+            db.close();
+        });
+}
+
+function readAndWriteFile(outputDirectory, dbGet, index) {
+    return dbGet('SELECT * FROM media LIMIT 1 OFFSET ?', index)
```
tfili: Running a new query for each row isn't efficient. We should do a select in chunks. Maybe like run the query to grab in 10,000 row chunks. You want to be careful with the gzipping and file writing as to not thrash disk IO.

I've done something similar like this:
```
var offset = 0;
var limit = 10000; // Or some reasonable limit
var asyncLimit = ??; // Number of cores or whatever
var processChunk = function() {
    return dbAll('SELECT * FROM media LIMIT ? OFFSET ?', limit, offset).then(function(rows) {
        if (rows.length === 0) {
            // No more rows left
            return Promise.resolve();
        }
        
        return Promise.map(rows, function(row) {
            // TODO: Handle gzipping/file writing
            ++offset;
        }, {concurrency: asyncLimit})
            .then(function () {
                return processChunk();
            });
    });
};

return processChunk();
```

lilleyse:
Updated. Thanks @tfili that code snippet was very helpful. The command is pretty fast now.

shehzan10:
LGTM.

shehzan10:
@lilleyse Do you think it would be good to deprecate instead of break this immediately? Probably have `tileset2sqlite3` call `tilesetToDatabase` internally. This would be useful in development as switching between latest master and npm release will not cause any breakage.

lilleyse:
This early in development, I think its fine to have the occasional breaking breaking change as long as it is documented - which means I should document it

lilleyse:
I'll add that and document the rest of the changes since the first release in a bit.

lilleyse:
Ready now.

PR #68, opened 2017-03-29
Update b3dm format and add new deprecated tilesets

lilleyse:
For https://github.com/AnalyticalGraphicsInc/cesium/pull/5165

### DIFF
--- a/generator/lib/createB3dm.js
+++ b/generator/lib/createB3dm.js
@@ -29,7 +29,8 @@ function createB3dm(options) {
     var featureTableJson = defaultValue(options.featureTableJson, defaultFeatureTable);
     var batchLength = featureTableJson.BATCH_LENGTH;
 
-    var featureTableJsonBuffer = getJsonBufferPadded(featureTableJson);
+    var headerByteLength = 28;
+    var featureTableJsonBuffer = getJsonBufferPadded(featureTableJson, headerByteLength);
     var featureTableBinary = getBufferPadded(options.featureTableBinary);
     var batchTableJsonBuffer = getJsonBufferPadded(options.batchTableJson);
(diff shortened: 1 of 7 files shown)


pjcozzi:
@shehzan10 can you please review and merge when you are available?

shehzan10:
I'm not entirely convinced that using `deprecated1` and `deprecated2` is a good. idea. I don't think there is enough intuitive distinction between then two.

One possible was to rename these are `options.header20Bytes` and `options.header24Bytes`. Assuming there is no versioning to the deprecated spec, I think we should use something more intuitive.

But otherwise looks good to me.

lilleyse:
Added some updates here.

* I realized the b3dm padding was not completely correct and all the glb's will start on a four byte boundary instead of eight. Doesn't really matter since the glbs here don't require it. Also not worth opening another data PR into Cesium for just that.
* Fixed issue where the `createBatchTableBinary` flag was being ignored for i3dm. Affects just one tileset, also not a huge deal.

> I'm not entirely convinced that using deprecated1 and deprecated2 is a good. idea. I don't think there is enough intuitive distinction between then two.

I think this is ok for now. This does represent the versions over time, so its easy for me to remember at least and isn't too wordy.

generator/lib/createB3dm.js
```diff
@@ -13,43 +13,63 @@ module.exports = createB3dm;
  * @param {Object} options An object with the following properties:
  * @param {Buffer} options.glb The binary glTF buffer.
  * @param {Number} [options.batchLength] The number of features in the tile.
```
shehzan10:
```
options.batchLength
```
This option should be removed, right?

lilleyse: Yup, removed.

PR #70, opened 2017-04-14
Clean up 3d-tiles-tools project and dependencies

mramato:
1. Update all package.json dependencies and switch to caret versioning.
2. Switched to `klaw`, which used to be bundled with `fs-extra` but they removed it and directed users to use it directly.
2. Clean up Webstorm Project, which was severely out of date.
3. Fix coverage.


mramato:
Travis is going to fail because it was enabled without having a `.travis.xml` set up in the root folder.  This will be fixed in a separate PR (probably by someone else).

tools/gulpfile.js
```diff
@@ -72,7 +72,11 @@ gulp.task('coverage', function () {
         ' cover' +
         ' --include-all-sources' +
         ' --dir coverage' +
-        ' -x "bin/** -x doc/** -x specs/** -x coverage/** -x index.js -x gulpfile.js"' +
+        ' -x "doc/**"' +
+        ' -x "specs/**"' +
+        ' -x "coverage/**"' +
+        ' -x "gulpfile.js"' +
+        ' -x "index.js"' +
```
shehzan10: Won't skipping the `bin` directory lower the coverage stats?

shehzan10:
> Travis is going to fail because it was enabled without having a .travis.xml set up in the root folder. This will be fixed in a separate PR (probably by someone else).

I think we can pretty much copy the entire [gltf-pipeline/.travis.yml](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/.travis.yml).

tools/gulpfile.js
mramato: We aren't skipping the `bin` directory.  And bin files can be programmatically unit tested just like the rest of the library so there's no reason to exclude them.

mramato:
> I think we can pretty much copy the entire gltf-pipeline/.travis.yml.

It's not that simple.  Since we have 3 separate npm packages in this repository, we need a top-level travis.yml that knows about each one and makes it easy to maintain and separate what each one does.  It's not complicated, but definitely worth a separate PR to implement.

mramato:
> Moved klaw to dependencies

Thanks, totally missed that.

lilleyse:
Thanks, looks good. I'll open an issue for fixing travis.

lilleyse:
Published to npm.

mramato:
Awesome, thanks.

PR #78, opened 2017-05-24
Fix translucent/opaque tileset techniques

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/cesium/pull/5308#discussion_r117840434

I had to do a similar thing in `obj2gltf` due to the inconsistencies of the `KHR_materials_common` extension.


lilleyse:
Sorry about the auto-close.

PR #84, opened 2017-06-21
Combine tilesets

lilleyse:
Adds the ability to combine external tilesets into a single tileset. It's been parked on this branch for a while and needed some small updates.

This PR also cleans up the tileset walking code a bit.

@shehzan10 can you review when you have a chance? No rush, I just want to get this out there so I can write another stage off of these changes.


tools/lib/isTileset.js
```diff
@@ -0,0 +1,12 @@
...
+var path = require('path');
+
+module.exports = isTileset;
+
+/**
+ * @private
+ */
+function isTileset(file) {
+    var extension = path.extname(file);
+    return extension === '.json';
```
shehzan10: Would it be more assertive to check the contents of the file for `asset`, `geometricError` and `root` properties which are all required for tileset.json?

The way it currently is, there is no different between `isJson` and `isTileset`.

tools/specs/lib/combineTilesetSpec.js
```diff
@@ -0,0 +1,186 @@
...
+            .then(function() {
+                return getNumberOfTilesets(combinedDirectory);
+            })
+            .then(function(numberOfTilesets) {
+                expect(numberOfTilesets).toBe(1);
+                return fsExtra.readFile(combinedJson, 'utf8');
+            })
+            .then(function(contents) {
+                var matches = getContentUrls(contents);
+                expect(matches).toEqual(['parent.b3dm', 'tileset3/ll.b3dm', 'lr.b3dm', 'ur.b3dm', 'ul.b3dm']);
```
shehzan10: What happens if there is an external tileset and more than one tileset.json refer to a single b3dm file? Will the copy function copy it to multiple directories?

Also, by extension, how will i3dm and the external glTF files for those be handled?

tools/lib/isTileset.js
lilleyse: Whoops, realized I'm not even using that file.

tools/specs/lib/combineTilesetSpec.js
lilleyse: I think both those cases should work out. When the tilesets are combined all paths are made relative to the root tileset, so the b3dm file that is referenced multiple times will appear with the same path throughout.

All files are just copied over straight and the original directory structure is preserved. External glTF files should be ok.

lilleyse:
Updated.

pjcozzi:
Merging so other PRs are not stalled.

PR #86, opened 2017-06-27
Upgrade tileset stage

lilleyse:
Review after #84 

This upgrades older tilesets to the latest 3D Tiles version. I've been testing this on the NYC CityGML tileset.

* Sets `asset.version` to 1.0
* Uses uppercase `ADD` and `REPLACE`
* Uses `_BATCHID` semantic instead of `BATCHID` in the glTF
* Upgrades b3dm tiles that use the deprecated headers

Todo:

* [ ] Requires https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/272 before embedded glTFs are actually converted to 2.0 like the doc states.


pjcozzi:
@likangning93 can you please review this?

tools/lib/extractB3dm.js
```diff
@@ -86,3 +97,10 @@ function extractB3dm(b3dmBuffer) {
         glb : glbBuffer
     };
 }
+
+function alignGlb(buffer, byteOffset) {
+    if (byteOffset % 4 === 0) {
+        return buffer;
+    }
+    return Buffer.from(buffer);
```
likangning93: From what I can tell, this is just copying `buffer` in memory if `byteOffset` isn't aligned to 4 bytes. Is this so the b3dm buffer can be modified somewhere without damaging the glb? It might be worth documenting.

tools/lib/isGzippedFile.js
```diff
@@ -1,15 +1,24 @@
...
  * @private
  */
 function isGzippedFile(file) {
-    return fsExtra.readFile(file)
-        .then(function (buffer) {
-            return isGzipped(buffer);
+    return new Promise(function (resolve, reject) {
+        var readStream = fsExtra.createReadStream(file, readStreamOptions);
+        readStream.on('error', reject);
+        readStream.on('data', function(chunk) {
```
likangning93: This is just checking the first chunk right? Should this be `readStream.once('data' //...`? Also, I'm still not 100% with streams, but should this get manually closed after the first chunk is read?

tools/lib/optimizeGlb.js
```diff
@@ -54,4 +53,40 @@ function optimizeGlb(glbBuffer, options) {
...
+    var meshes = gltf.meshes;
+    for (var meshId in meshes) {
+        if (meshes.hasOwnProperty(meshId)) {
+            var primitives = meshes[meshId].primitives;
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; ++i) {
+                var attributes = primitives[i].attributes;
+                for (var semantic in attributes) {
+                    if (attributes.hasOwnProperty(semantic)) {
+                        if (semantic === 'BATCHID') {
```
likangning93: Is it a bad idea to directly look up `attributes.BATCHID`? If not, it might be a little cleaner.

tools/lib/optimizeGlb.js
```diff
@@ -54,4 +53,40 @@ function optimizeGlb(glbBuffer, options) {
...
+    }
+
+    var techniques = gltf.techniques;
+    for (var techniqueId in techniques) {
+        if (techniques.hasOwnProperty(techniqueId)) {
+            var parameters = techniques[techniqueId].parameters;
+            for (var parameterId in parameters) {
+                if (parameters.hasOwnProperty(parameterId)) {
+                    var parameter = parameters[parameterId];
+                    if (parameter.semantic === 'BATCHID') {
```
likangning93: Similar to above.

tools/specs/lib/upgradeTilesetSpec.js
```diff
@@ -0,0 +1,192 @@
...
+var upgradedDirectory = './specs/data/BatchedDeprecated-upgraded/';
+var upgradedJson = './specs/data/BatchedDeprecated-upgraded/tileset.json';
+var upgradedB3dm1 = './specs/data/BatchedDeprecated-upgraded/batchedDeprecated1.b3dm';
+var upgradedB3dm2 = './specs/data/BatchedDeprecated-upgraded/batchedDeprecated2.b3dm';
+
+describe('upgradeTileset', function() {
+    afterEach(function (done) {
+        Promise.all([
+            fsExtra.remove(upgradedDirectory),
+            fsExtra.remove(gzippedDirectory)
```
likangning93: since `upgradeTileset` takes a `write` callback, should we try to avoid filesystem changing operations in this spec?

tools/lib/optimizeGlb.js
lilleyse: I think this case is different.

tools/specs/lib/upgradeTilesetSpec.js
lilleyse: Yeah this needs to be fixed throughout the codebase. I'll write an issue.

lilleyse:
Addressed your comments @likangning93.

I also support for upgrading cmpt tiles in https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/86/commits/2b787f4eccaad80331ee725be4bb7f52287e3d27 and in the process took code from #4.

tools/lib/optimizeGlb.js
likangning93: ah right...

PR #88, opened 2017-06-29
Use nyc for coverage

omh1280:
Add [nyc](https://github.com/istanbuljs/nyc). to each of the three repositories for code coverage.

`nyc` is the new version of `istanbul`. (They retained the istanbul name for he actual coverage library, but nyc is now the command line tool to run it).  This happened a while ago and we've been on an old version of istanbul until now.

The `tools` tests don't run on my computer, but the other two repositories work and show the coverage successfully.


mramato:
Thanks.

PR #102, opened 2017-07-07
Cleared eslint issues

Rudraksha20:
@lilleyse cleared some eslint errors.

### DIFF
--- a/validator/specs/lib/validateTilesetSpec.js
+++ b/validator/specs/lib/validateTilesetSpec.js
@@ -67,7 +67,7 @@ describe('validateTileset', function() {
                 expect(message).toBe('Child has geometricError greater than parent');
             }), done).toResolve();
     });
-  
+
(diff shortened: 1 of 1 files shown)


validator/specs/lib/validateTilesetSpec.js
```diff
@@ -110,7 +110,7 @@ describe('validateTileset', function() {
         expect(validateTileset(tileset)
             .then(function(message) {
                 expect(message).toBe('Tileset must declare a version in its asset property');
-            }), done).toResolve();
+           }), done).toResolve();
```
lilleyse: Undo this change.

lilleyse:
Looks good besides that one comment.

validator/specs/lib/validateTilesetSpec.js
Rudraksha20: Done.

PR #103, opened 2017-07-07
Check for batch_ids to be less than batch_length

### LINKED ISSUE #99: Validator - Check that BATCH_ID values are correct
@rms13

i3dm and pnts tiles contain an optional `BATCH_ID` semantic in their feature table. If this property exists, its values should be less than `BATCH_LENGTH`.

To help parse the contents of a feature table, you can require Cesium's `Cesium3DTileFeatureTable` file similar to how `defined` and `defaultValue` are often required.

rms13:
Fixes #99.

Added a test in validatePntsSpec.js for the check.

### DIFF
--- a/validator/lib/validatePnts.js
+++ b/validator/lib/validatePnts.js
@@ -9,8 +9,6 @@ var featureTableSchema = require('../specs/data/schema/featureTable.schema.json'
 
 var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
-// var Cesium3DTileFeatureTable = Cesium.Cesium3DTileFeatureTable;
-// var ComponentDatatype = Cesium.ComponentDatatype;
 
 module.exports = validatePnts;
 
--- a/validator/lib/validatePnts.js
+++ b/validator/lib/validatePnts.js
@@ -196,6 +196,4 @@ function validatePnts(content) {
     if (defined(batchTableMessage)) {
         return batchTableMessage;
     }
-
-
 }
--- a/validator/lib/validatePnts.js
+++ b/validator/lib/validatePnts.js
@@ -9,6 +9,8 @@ var featureTableSchema = require('../specs/data/schema/featureTable.schema.json'
 
 var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
+var Cesium3DTileFeatureTable = Cesium.Cesium3DTileFeatureTable;
+var ComponentDatatype = Cesium.ComponentDatatype;
 
 module.exports = validatePnts;
 
@@ -182,8 +184,12 @@ function validatePnts(content) {
     }
 
     if (defined(featureTableJson.BATCH_ID)) {
-        if (featureTableJson.BATCH_ID.length > featureTableJson.BATCH_LENGTH) {
-            return 'Feature table property BATCH_LENGTH must be greater than total number of BATCH_IDs defined.';
+        var featureTable = new Cesium3DTileFeatureTable(featureTableJson, featureTableBinary);
+        var batchIds = featureTable.getPropertyArray('BATCH_ID', ComponentDatatype.UNSIGNED_SHORT, 1);
+        for (var i = 0; i < batchIds.length; i++) {
+             if (featureTableJson.BATCH_LENGTH <= batchIds[i]) {
+                return 'Feature table property BATCH_LENGTH must be greater than all the elements of BATCH_ID.';
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


validator/lib/validatePnts.js
```diff
@@ -181,6 +181,12 @@ function validatePnts(content) {
         return 'Feature table property BATCH_LENGTH must be less than or equal to POINTS_LENGTH.';
     }
 
+    if (defined(featureTableJson.BATCH_ID)) {
+        if (featureTableJson.BATCH_ID.length > featureTableJson.BATCH_LENGTH) {
+            return 'Feature table property BATCH_LENGTH must be greater than total number of BATCH_IDs defined.';
```
lilleyse: This isn't exactly the right thing to check.

Instead you should check that each value inside the `BATCH_ID` array is less than `BATCH_LENGTH`. This is where `Cesium3DTileFeatureTable.getPropertyArray` comes in, letting you get all the values and then looping over them.

rms13:
Corrected!

validator/lib/validatePnts.js
```diff
@@ -181,6 +183,16 @@ function validatePnts(content) {
         return 'Feature table property BATCH_LENGTH must be less than or equal to POINTS_LENGTH.';
     }
 
+    if (defined(featureTableJson.BATCH_ID)) {
+        var featureTable = new Cesium3DTileFeatureTable(featureTableJson, featureTableBinary);
+        var batchIds = featureTable.getPropertyArray('BATCH_ID', ComponentDatatype.UNSIGNED_SHORT, 1);
```
lilleyse: The `BATCH_ID` property is the one exception where it may declare a component type other than the default of `UNSIGNED_SHORT`.

Cesium has the same mistake which I'll address. The fixed code might look like:

```
var componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.BATCH_ID.componentType, 'UNSIGNED_SHORT'));
var batchIds = featureTable.getPropertyArray('BATCH_ID', componentDatatype, 1);

File: validator/lib/validatePnts.js [hunk 3]
```diff
@@ -181,6 +183,16 @@ function validatePnts(content) {
...
+        for (var i = 0; i < batchIds.length; i++) {
```
lilleyse: As a style suggestion, we usually factor out the array length into its own variable, like:

```
var length = batchIds.length;
for (var i = 0; i < length ; i++) {
    ...
}
```

This is generally good practice in JavaScript for performance intensive areas. Even though this area isn't one of those we still prefer to write for loops in this way.

validator/lib/validatePnts.js
```diff
@@ -181,6 +183,16 @@ function validatePnts(content) {
...
+             if (featureTableJson.BATCH_LENGTH <= batchIds[i]) {
```
lilleyse: Reversing the condition is slightly clearer:

`if (batchIds[i] >= featureTableJson.BATCH_LENGTH) {`

Also tweak the description to match.

validator/specs/lib/validatePntsSpec.js
```diff
@@ -136,6 +136,18 @@ describe('validate pnts', function() {
         expect(validatePnts(pnts)).toBe('Feature table property BATCH_LENGTH must be less than or equal to POINTS_LENGTH.');
     });
 
+    it('returns error message if any BATCH_ID is greater than BATCH_LENGTH: ', function() {
+        var pnts = createPnts({
+            featureTableJson : {
+                POINTS_LENGTH : 1,
+                POSITION : [0, 0, 0],
+                BATCH_ID : [2],
+                BATCH_LENGTH : 1
```
lilleyse: For a slightly stronger test consider testing with `POINTS_LENGTH`/`BATCH_LENGTH` being greater than 1. This will guarantee that the loop iteration is working as it should.

lilleyse: Also add a second test that does the same thing but using a batch table binary.

lilleyse:
Thanks for the update @rms13.

Make sure to always leave a quick message when you've pushed new code that is ready to review, that way I get notified of the changes.

lilleyse:
The new code changes look great!

Just add a second test that utilizes a batch table binary. This will help test the robustness of `Cesium3DTileFeatureTable`: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/103/files/bcb1fa969f43723cc739bb51e3ae829f87501163#r126230439

lilleyse:
Correction: create a test using a feature table binary, not a batch table binary as I suggested.

rms13:
@lilleyse 
Test for feature table binary added.

lilleyse:
Nice, works great!

Issue #98, opened 2017-07-06
Validator - validate glb


lilleyse:
@Rudraksha20

After the small tileset.json fixes are squared away we will work on validating the glTF models that are embedded within b3dm and i3dm tiles. This will be a multi-part issue that we'll add on to as we go.

Open a PR for each checkbox.

* [x] Check that the embedded glb is version 2.0. If not return an error message.
* [ ] Check that the glb meets the glTF 2.0 schema. We'll need to use the [glTF validator](https://github.com/KhronosGroup/glTF-Validator) here, probably the `dev` branch. Most likely the glTF validator will need to be called via a child process.

javagl:
As mentioned in https://github.com/CesiumGS/3d-tiles-validator/pull/222#issuecomment-1265905466 : The updated validator will validate glTF assets that are found as tile content, using the glTF validator.

The issues that are found by the glTF validator will be the `causes` of a single `CONTENT_VALIDATION_ERROR` that is created in these cases, like in this example:
```
Validation result:
{
  "date": "2022-10-04T15:03:45.676Z",
  "numErrors": 1,
  "numWarnings": 0,
  "issues": [
    {
      "type": "CONTENT_VALIDATION_ERROR",
      "path": "tiles/glTF/TriangleGlbWithErrors/TriangleGlbWithInvalidLength.glb",
      "message": "Content tiles/glTF/TriangleGlbWithErrors/TriangleGlbWithInvalidLength.glb caused validation errors",
      "severity": "ERROR",
      "causes": [
        {
          "type": "CONTENT_VALIDATION_ERROR",
          "message": "Chunk (0x004e4942) length (44) does not fit total GLB length.",
          "severity": "ERROR"
        },
        {
          "type": "CONTENT_VALIDATION_ERROR",
          "message": "Declared length (680) does not match GLB length (684).",
          "severity": "ERROR"
        }
      ]
    }
  ]
}
```

PR #104, opened 2017-07-07
validateGlb version check

Rudraksha20:
Worked on issue #98. Updated `validateGlb` to check for version, added spec `validateGlbSpec`, also updated `specUtility`.

### DIFF
--- a/validator/specs/lib/specUtility.js
+++ b/validator/specs/lib/specUtility.js
@@ -38,7 +38,7 @@ function createB3dm(options) {
         batchTableBinary = Buffer.concat([batchTableJsonBuffer, Buffer.alloc(1)]);
     }
     if (options.unalignedByteLength) {
-        glb = Buffer.concat([glb, createGlb()]);
+        glb = Buffer.concat([glb, Buffer.alloc(1)]);
     }
 
     var featureTableJsonByteLength = featureTableJsonBuffer.length;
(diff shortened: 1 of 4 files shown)


validator/specs/lib/specUtility.js
```diff
@@ -37,7 +38,7 @@ function createB3dm(options) {
         batchTableBinary = Buffer.concat([batchTableJsonBuffer, Buffer.alloc(1)]);
     }
     if (options.unalignedByteLength) {
-        glb = Buffer.concat([glb, Buffer.alloc(1)]);
+        glb = Buffer.concat([glb, createGlb()]);
```
lilleyse: This change isn't needed, this area is purposefully adding an extra byte to throw off byte alignment (tested in `validateCmpt`.

I believe the tests are still passing because the glb buffer is 12 bytes and therefore still doesn't meet the 8-byte alignment requirement.

validator/specs/lib/specUtility.js
```diff
@@ -114,7 +115,7 @@ function createI3dm(options) {
...
```
lilleyse: Same comment here.

validator/specs/lib/specUtility.js
```diff
@@ -231,3 +232,11 @@ function getJsonBufferPadded(json, byteOffset) {
 
     return Buffer.from(string);
 }
+
+function createGlb() {
+    var Glb = Buffer.alloc(12);
```
lilleyse: Use lower case `glb` for variable names.

Rudraksha20: Updated.

lilleyse:
Thanks!

Issue #25, opened 2016-11-01
Validator - Bounding Volume Spatial Coherence


lilleyse:
For #1 
@sumitshyamsukha 

Let's do this one now:

> tile.content.boundingVolume, when defined, is completely inside tile.boundingVolume (use Cesium's functions)

There are three bounding volume types: region, box, and sphere. They are defined here: https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/schema/boundingVolume.schema.json

If a tile has a content bounding volume, check that it is inside it's bounding volume. To visualize what a bounding volume / content bounding volume looks like, go to [the 3D Tiles Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles.html&label=Showcases) when running the 3d-tiles branch in Cesium and check out the `BV On/Off` and `Contents BV On/Off` buttons.

We'll need support the following checks:
* [ ] Region inside region
* [ ] Box inside region
* [ ] Sphere inside region
* [ ] Sphere inside sphere
* [ ] Region inside sphere
* [ ] Box inside sphere
* [ ] Box inside box
* [ ] Region inside box
* [ ] Sphere inside box

I'll update with ideas for solving some of these, but to start region inside region should be pretty straightforward. Just need to check that the west/south/east/north/min/max are within the other's values.

pjcozzi:
> I'll update with ideas for solving some of these, but to start region inside region should be pretty straightforward. Just need to check that the west/south/east/north/min/max are within the other's values

Please add functions to Cesium as needed.

sumitshyamsukha:
@lilleyse Would you like me to add this functionality to the `validateTileset.js` file along with the geometricError validation, or should I create a new one for all of the bounding volume validation?

lilleyse:
Yes add it to the existing traversal in `validateTileset`.

lilleyse:
@rms13 let's finish the most important ones here, starting with `Box inside box`.

For this it may help to reference Cesium's `OrientedBoundingBox` class and look at how it's created from the `box` array here: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/Cesium3DTile.js#L828.

[...]

PR #105, opened 2017-07-10
Check for box type content bounding volume inside tile with box type bounding volume

rms13:
Issue #25 - Box inside box

@lilleyse, According to the [3D tiles readme](https://github.com/AnalyticalGraphicsInc/3d-tiles#tile-transform), tile.transform applies to both - the tile's bounding box, and its content's bounding box. So, to check if one is inside the other, the transform should not affect the result. 

So, I have neglected transform for the check. I am converting the half axes representation of the boxes to a diagonal representation, and comparing both the boxes directly, assuming they have the same orientation.

Also, there was a minor spelling error in sphere inside sphere check, which I have corrected here.

### DIFF
--- a/validator/lib/utility.js
+++ b/validator/lib/utility.js
@@ -96,10 +96,9 @@ function boxInsideBox(boxInner, boxOuter) {
     cube[6] = new Cartesian3(1, 1, 1);
     cube[7] = new Cartesian3(1, 1, -1);
 
-    var i = 0;
     var transformInnerInverse = Matrix4.inverse(transformOuter, transformOuter);
     var EPSILON8 = CesiumMath.EPSILON8;
-    for (i = 0; i < 8; i++) {
+    for (var i = 0; i < 8; i++) {
         cube[i] = Matrix4.multiplyByPoint(transformInner, cube[i], cube[i]);
         cube[i] = Matrix4.multiplyByPoint(transformInnerInverse, cube[i], cube[i]);
         var min = Cartesian3.minimumComponent(cube[i]);
--- a/validator/lib/utility.js
+++ b/validator/lib/utility.js
@@ -2,6 +2,7 @@
 var Cesium = require('cesium');
 
 var Cartesian3 = Cesium.Cartesian3;
+var CesiumMath = Cesium.Math;
 var Matrix3 = Cesium.Matrix3;
 var Matrix4 = Cesium.Matrix4;
 
@@ -105,8 +106,14 @@ function boxInsideBox(boxInner, boxOuter) {
         cube[i] = Matrix4.multiplyByPoint(transformInnerInverse,cube[i],cube[i]);
     }
 
+    var EPSILON8 = CesiumMath.EPSILON8;
     for (i = 0; i < 8; i++) {
-        if (cube[i].x < 0 || cube[i].x > 1 || cube[i].y < 0 || cube[i].y > 1 || cube[i].z < 0 || cube[i].z > 1) {
+        if (cube[i].x < 0 - EPSILON8 ||
+            cube[i].x > 1 + EPSILON8 ||
+            cube[i].y < 0 - EPSILON8 ||
+            cube[i].y > 1 + EPSILON8 ||
+            cube[i].z < 0 - EPSILON8 ||
+            cube[i].z > 1 + EPSILON8) {
             return false;
         }
     }
--- a/validator/lib/utility.js
+++ b/validator/lib/utility.js
@@ -77,17 +77,14 @@ var scratchInnerHalfAxes = new Matrix3();
 var scratchOuterHalfAxes = new Matrix3();
 
 function boxInsideBox(boxInner, boxOuter) {
-    // Compute inner box..
     var centerInner = Cartesian3.fromElements(boxInner[0], boxInner[1], boxInner[2], scratchInnerCenter);
     var halfAxesInner = Matrix3.fromArray(boxInner, 3, scratchInnerHalfAxes);
     var transformInner = Matrix4.fromRotationTranslation(halfAxesInner,centerInner);
 
-    // Compute outer box..
     var centerOuter = Cartesian3.fromElements(boxOuter[0], boxOuter[1], boxOuter[2], scratchOuterCenter);
     var halfAxesOuter = Matrix3.fromArray(boxOuter, 3, scratchOuterHalfAxes);
     var transformOuter = Matrix4.fromRotationTranslation(halfAxesOuter,centerOuter);
 
-    // Create a unit cube 0,0,0 to 1,1,1..
     var cube = new Array(8);
     cube[0] = new Cartesian3(0, 0, 0);
     cube[1] = new Cartesian3(0, 0, 1);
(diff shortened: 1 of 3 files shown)


lilleyse:
Yeah fine to ignore the transform for now. This may matter later once once we do bounding volume checks with a child against its parent.

> assuming they have the same orientation

While this will usually be true in practice, the validator should be able to test against any arbitrary obb's.

rms13:
@lilleyse 
So the half-axes represent the transforms of individual bounding box, which can be tile or content.
The transform specified in the tile will never contribute to relative orientation of the tile and its content, but the half-axes will (for box type BB). 

So, I have changed the previous implementation of box inside box check in utility.js to take into consideration the orientation defined by the half-axes, for both the tile and the content.

validator/lib/utility.js
```diff
@@ -58,14 +62,60 @@ function regionInsideRegion(regionInner, regionOuter) {
...
     return distance <= (radiusOuter - radiusInner);
 }
+
+var scratchInnerHalfAxes = new Matrix3();
+var scratchOuterHalfAxes = new Matrix3();
+
+function boxInsideBox(boxInner, boxOuter) {
+    var centerInner = Cartesian3.fromElements(boxInner[0], boxInner[1], boxInner[2], scratchInnerCenter);
+    var halfAxesInner = Matrix3.fromArray(boxInner, 3, scratchInnerHalfAxes);
+    var transformInner = Matrix4.fromRotationTranslation(halfAxesInner,centerInner);
```
lilleyse: Style: add space after comma - `(halfAxesInner, centerInner);`

lilleyse: Fix throughout.

validator/lib/validateTileset.js
```diff
@@ -104,6 +108,10 @@ function validateTileHierarchy(root, tilesetDirectory) {
             if (defined(contentSphere) && defined(tileSphere) && !sphereInsideSphere(contentSphere, tileSphere)) {
                 return 'content sphere [' + contentSphere + '] is not within tile sphere + [' + tileSphere + ']';
             }
+
+            if (defined(contentBox) && defined(tileBox) && !boxInsideBox(contentBox, tileBox)) {
+                return 'content box is not within tile box';
```
lilleyse: Also include the `contentBox` and `tileBox` in the message, similar to the messages right above.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -131,6 +164,14 @@ describe('validateTileset', function() {
...
     });
 
+    it('returns error message when a content\'s box type boundingVolume is not within it\'s tile\'s box type boundingVolume', function(done) {
+        var tileset = clone(sampleTileset2, true);
+        expect(validateTileset(tileset)
+            .then(function(message) {
+                expect(message).toBe('content box is not within tile box');
+            }), done).toResolve();
+    });
+
```
lilleyse: Can you also add a second test with two OBBs that are not aligned? Also add a test that checks a valid tileset.

It may be worth creating a helper function to generate the tileset based on a provided `boundingVolume` and `contentBoundingVolume` that all these tests can use (and https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/109/files too).

validator/lib/utility.js
```diff
@@ -58,14 +62,60 @@ function regionInsideRegion(regionInner, regionOuter) {
...
+
+    var cube = new Array(8);
+    cube[0] = new Cartesian3(0, 0, 0);
+    cube[1] = new Cartesian3(0, 0, 1);
+    cube[2] = new Cartesian3(1, 0, 1);
+    cube[3] = new Cartesian3(1, 0, 0);
+    cube[4] = new Cartesian3(0, 1, 0);
+    cube[5] = new Cartesian3(0, 1, 1);
+    cube[6] = new Cartesian3(1, 1, 1);
+    cube[7] = new Cartesian3(1, 1, 0);
```
lilleyse: After looking at the doc here: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/OrientedBoundingBox.js#L42-L44, I think the 0's need to be -1 so the box is a 2x2x2 box.

rms13: I think it should be fine as long as I am comparing the resultant values to the right numbers. Still I am changing these to -1 to maintain consistency.

rms13:
@lilleyse Made the changes. Please review.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -131,10 +131,102 @@ describe('validateTileset', function() {
...
+            ]
+        };
+        var tileset = createSampleTileset(tileBoundingVolume, contentBoundingVolume);
+        expect(validateTileset(tileset)
+            .then(function(message) {
+                expect(message).toBe('content box [' + contentBoundingVolume.box + '] is not within tile box [' + tileBoundingVolume.box + ']');
+            }), done).toResolve();
+    });
+
+    it('returns error message when a content\'s box type boundingVolume is not within it\'s tile\'s box type boundingVolume [valid bounding boxes]', function(done) {
```
lilleyse: Fix description, since this is a case where it is valid.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -131,10 +131,102 @@ describe('validateTileset', function() {
...
+        var tileBoundingVolume = {
+            box: [0, 0, 0,
```
lilleyse: Be consistent with formatting of all the `box` areas in the file.

validator/lib/utility.js
```diff
@@ -58,14 +62,60 @@ function regionInsideRegion(regionInner, regionOuter) {
...
+    var EPSILON8 = CesiumMath.EPSILON8;
+    for (i = 0; i < 8; i++) {
+        if (cube[i].x < -1 - EPSILON8 ||
+            cube[i].x > 1 + EPSILON8 ||
+            cube[i].y < -1 - EPSILON8 ||
+            cube[i].y > 1 + EPSILON8 ||
+            cube[i].z < -1 - EPSILON8 ||
+            cube[i].z > 1 + EPSILON8) {
+            return false;
+        }
```
lilleyse: Can these 3 for loops be combined? The last one can be condensed a bit by using `Cartesian3.minimumComponent` and `Cartesian3.maximumComponent`.

rms13:
@lilleyse 
Updated.

validator/lib/utility.js
```diff
@@ -58,14 +62,53 @@ function regionInsideRegion(regionInner, regionOuter) {
...
+    cube[6] = new Cartesian3(1, 1, 1);
+    cube[7] = new Cartesian3(1, 1, -1);
+
+    var i = 0;
+    var transformInnerInverse;
+    var min, max;
+    var EPSILON8 = CesiumMath.EPSILON8;
+    for (i = 0; i < 8; i++) {
+        cube[i] = Matrix4.multiplyByPoint(transformInner, cube[i], cube[i]);
+        transformInnerInverse = Matrix4.inverse(transformOuter, transformOuter);
```
lilleyse: `transformInnerInverse` can be computed outside of the loop.

validator/lib/utility.js
```diff
@@ -58,14 +62,53 @@ function regionInsideRegion(regionInner, regionOuter) {
...
+    cube[4] = new Cartesian3(-1, 1, -1);
+    cube[5] = new Cartesian3(-1, 1, 1);
+    cube[6] = new Cartesian3(1, 1, 1);
+    cube[7] = new Cartesian3(1, 1, -1);
+
+    var i = 0;
+    var transformInnerInverse;
+    var min, max;
+    var EPSILON8 = CesiumMath.EPSILON8;
+    for (i = 0; i < 8; i++) {
```
lilleyse: Now that there is one for loop, it is fine to do `for (var i = 0`.

Similarly `min` and `max` do not need to be declared outside of the loop.

lilleyse:
Merging #109 introduced some merge conflicts here.

rms13:
@lilleyse Should merge now.

lilleyse:
👍

PR #106, opened 2017-07-12
Running GLTF as ext process

Rudraksha20:
@lilleyse 

Issue #98 

Note: I have included the source code for GLTF_validator and two folders `exec` which has the `gilt_validator` executable that will be replaced and `glb_sample` that has the `Box.glb` sample.

How to run GLTF as an ext process:

Prerequisites:

1. Download and install `Dart SDK` for your platform. [https://www.dartlang.org/install](url)

2. Add Dart SDK `bin` folder to your PATH. [https://www.dartlang.org/tools/pub/installing](url)

3. Add packages `bin` folder to your PATH (`~/.pub-cache/bin` for Linux and Mac; `%APPDATA%\Pub\Cache\bin` for Windows).

glTF-Validator:

1. From the repository root folder (go in folder `GLTF/glTF-Validator-master/`), run `pub get` to get dependencies.

2. Run `pub global activate --source path ./` to add gltf_validator executable to your PATH.

3. On a mac and linux go to `.pub_cache` in your home directory and there you will find the `gltf_validator` executable. Take this executable file and replace the executable in `GLTF/exec/` located in your 3D-Tiles validator with it.
    For Windows the executable is in the folder `AppData\Roaming\Pub\Cache` do the same as above.

Running the program:

1. Run `npm run test` from the `terminal`. Make sure you are in the `3D-tiles-tools/validator` directory.

Once the above test works you should see all the test cases pass as well as an error will be printed that the version of Gltf is wrong, which is to be expected.


pjcozzi:
@lilleyse throughout reviews, please make sure we are tracking licenses for third-party libraries.

validator/lib/validateGlb.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = validateGlb;
-
+var cp = require('child_process');
```
lilleyse: Style: avoid abbreviations. Also place this above `module.exports`.

validator/lib/validateGlb.js
```diff
@@ -11,7 +11,14 @@ module.exports = validateGlb;
...
+        if(err) {console.error(err);}    
+        console.log('stdout', stdout);
+        console.log('stderr', stderr);
+    });
+
     if (version !== 2) {
         return 'Invalid Glb version: ' + version + '. Version must be 2.';
     }
 }
+
```
lilleyse: Remove empty line.

validator/lib/validateGlb.js
```diff
@@ -11,7 +11,14 @@ module.exports = validateGlb;
 function validateGlb(glb) {
     var version = glb.readUInt32LE(4);
 
+    cp.execFile('./GLTF/exec/gltf_validator', ['./GLTF/glb_sample/Box.glb'], function(err, stdout, stderr){
+        if(err) {console.error(err);}    
+        console.log('stdout', stdout);
+        console.log('stderr', stderr);
+    });
```
lilleyse: This should go after the version check. 

While Node functions take callbacks, we prefer to use Promises. Check out https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/compressTexture.js#L207-L221 for an example of wrapping a Promise around a callback.

`validateGlb` will need to return a promise, and likewise so will `validateB3dm`, which means most areas that call `validateB3dm` will need to expect a promise, and not just a message.

I had some problems running this on Windows initially. `gltf_validator` is called `gltf_validator.bat`. I think `spawn` will work better to ignore the extension. Also its probably best to reference an absolute path for the validator exe - `var gltfValidatorPath= path.join(__dirname, '../bin/gltf_validator);`

Does the gltf validator accept a buffer as input, or does it only take files? If only files, the `glb` passed into `validateGlb` will need to be saved to a temporary file before being passed in. Check out https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/getTempDirectory.js for a part-way example.

lilleyse:
`./GLTF/exec/gltf_validator` - in the past we put exes like this in the 'bin' folder.

lilleyse:
glTF validation should be optional since the setup required is a bit involved. `validateGlb` should recognize when the gltf-validator is not setup, maybe by checking if the exe actually exists in the bin folder. If it is setup, run through the validator, otherwise don't.

lilleyse:
Put any sample models in `specs/data/` folder instead.

lilleyse:
Add these notes to a section in the `README`.

lilleyse:
It would be cleaner to include gltf validator as a git submodule.

lilleyse:
Please add gltf validator to LICENSE.md under the `Third-Party Code` section. Reference: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/blob/master/samples-generator/LICENSE.md#third-party-code

lilleyse:
Add some tests to `validateGlbSpec`

validator/lib/validateGlb.js
```diff
@@ -11,7 +11,14 @@ module.exports = validateGlb;
 function validateGlb(glb) {
     var version = glb.readUInt32LE(4);
 
+    cp.execFile('./GLTF/exec/gltf_validator', ['./GLTF/glb_sample/Box.glb'], function(err, stdout, stderr){
+        if(err) {console.error(err);}    
+        console.log('stdout', stdout);
+        console.log('stderr', stderr);
```
lilleyse: Once promises are set up, this should not call `console.log`, instead the promise should be resolved with the output message.

validator/lib/validateGlb.js
```diff
@@ -11,7 +11,14 @@ module.exports = validateGlb;
 function validateGlb(glb) {
     var version = glb.readUInt32LE(4);
 
+    cp.execFile('./GLTF/exec/gltf_validator', ['./GLTF/glb_sample/Box.glb'], function(err, stdout, stderr){
+        if(err) {console.error(err);}
```
lilleyse: Use Cesium's `defined` function here.

mramato:
> glTF validation should be optional since the setup required is a bit involved.

Have we looked into using `dart2js` to compile the module into a Node JS library that we can just suck in like any other Node module?  `dart2js` is an official part of the dart project and how most people ship Dart code as JS.

pjcozzi:
There's been talk of using `dartjs`, but it is not scoped for this project unless it turns out to be trivial.  Feel free to discuss with @lexaknyazev who plans to investigate at some point.

Rudraksha20:
@lilleyse I have updated the files, they have the following changes as suggested:

- Style improvements.
- Exec and data files in the bin and specs folder respectively.
- GLTF validator runs only if the executable exists.
- Added test to validateGLB spec.
- Updated LICENSE.md
- Updated README.md

Also, the current version fails for the tests where the I3dm, b3dm, pnts and tile files call `validateGLB` and expect undefined behavior. I have tried to remove those errors, but could I get some help debugging them?

Issue #107, opened 2017-07-12
Validator - report line number of error in tileset.json


lilleyse:
@Rudraksha20

It would be helpful if `validateTileset` could report the line number of the JSON at which the error occurred. Try out some ideas to see if this is actually doable.

javagl:
In the updated validator, the issues will contain the "JSON paath" pointing to the element that caused the issue.

PR #108, opened 2017-07-13
Tiles meaningful debugging output

Rudraksha20:
@lilleyse 
Issue #107, validator outputs the part of the Json object where the error occurred. The output object only includes its properties but does not include children.


Rudraksha20:
@lilleyse Updated the tilesetValidator to output line number or Json for the error reporting.

validator/lib/validateTileset.js
```diff
@@ -98,16 +154,25 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+                console.log(string_json);
                 return 'content sphere [' + contentSphere + '] is not within tile sphere + [' + tileSphere + ']';
             }
         }
 
         if (defined(tile.refine)) {
             if (tile.refine !== 'ADD' && tile.refine !== 'REPLACE') {
+                var string_json = JSON.stringify(tile, null, 4);
+                console.log('Error in tile:');
+                console.log(string_json);
```
lilleyse: Create a helper function for all of these. It will take the tile and the error message (e.g. `'Refine property in tile must have either "ADD" or "REPLACE" as its value.'`) and return an error message with the string json appended.

validator/lib/validateTileset.js
```diff
@@ -98,16 +154,25 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+                var string_json = JSON.stringify(tile.content, null, 4);
+                console.log('Error in tile:');
+                console.log(string_json);
                 return 'content sphere [' + contentSphere + '] is not within tile sphere + [' + tileSphere + ']';
             }
         }
 
         if (defined(tile.refine)) {
             if (tile.refine !== 'ADD' && tile.refine !== 'REPLACE') {
+                var string_json = JSON.stringify(tile, null, 4);
```
lilleyse: We stick to camel case: `stringJson`. Also we prefer `undefined` to `null`.

validator/lib/validateTileset.js
lilleyse: For me the children array is printed in its entirety, so it would be best if this helper function also deleted `children`.

Rudraksha20:
@lilleyse Updated with recommendations!

validator/package.json
```diff
@@ -24,7 +24,8 @@
   "dependencies": {
     "bluebird": "^3.5.0",
     "cesium": "^1.34",
-    "fs-extra": "^3.0.1"
+    "fs-extra": "^3.0.1",
+    "linenumber": "^1.0.0"
```
lilleyse: Remove from `package.json`.

validator/lib/validateTileset.js
```diff
@@ -33,28 +33,70 @@ function validateTileset(tileset, tilesetDirectory) {
...
 function validateTopLevel(tileset) {
     if (!defined(tileset.geometricError)) {
+        var temp_tileset = tileset;
+        if (defined(temp_tileset.root)) {
+            delete temp_tileset.root;
+        }
+        var string_json = JSON.stringify(temp_tileset, undefined, 4);
+        console.log('Error in tile:');
+        console.log(string_json);
         return 'Tileset must declare its geometricError as a top-level property.';
```
lilleyse: Let's not print the JSON for the top-level properties, since the JSON doesn't add much value to the error message.

validator/lib/validateTileset.js
```diff
@@ -76,14 +118,35 @@ function validateTileHierarchy(root, tilesetDirectory) {
         var content = tile.content;
 
         if (!defined(tile.geometricError)) {
+            var temp_json = tile;
+            if (defined(temp_json.children)) {
+                delete temp_json.children;
+            }
+            var string_json = JSON.stringify(temp_json, undefined, 4);
+            console.log('Error in tile:');
+            console.log(string_json);
```
lilleyse: Make a helper function for all these - see this comment: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/108#discussion_r127732457.

Avoid using `console.log` for error messages. Everything should be contained in the message.

Creating `temp_json` isn't actually doing anything since `temp_json` and `tile` still refer to the same object. It's fine not to create the temp variable since the validation will exit right after, so modifying the json doesn't matter.

validator/lib/validateTileset.js
```diff
@@ -76,14 +118,35 @@ function validateTileHierarchy(root, tilesetDirectory) {
         var content = tile.content;
 
         if (!defined(tile.geometricError)) {
+            var temp_json = tile;
+            if (defined(temp_json.children)) {
+                delete temp_json.children;
+            }
```
lilleyse:
```
if (defined(temp_json.children)) {
    delete temp_json.children;
}
```

can be condensed to just `delete temp_json.children`

validator/lib/validateTileset.js
```diff
@@ -76,14 +118,35 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            var string_json = JSON.stringify(temp_json, undefined, 4);
```
lilleyse: Style: we use camel case throughout: `stringJson`

Rudraksha20:
@lilleyse Updated!

validator/GLTF/glTF-Validator-master/.packages
```diff
@@ -0,0 +1,54 @@
+# Generated by pub on 2017-08-09 11:40:43.623996.
```
lilleyse: It looks like some files got added to this PR accidentally.

validator/lib/validateTileset.js
```diff
@@ -36,19 +36,19 @@ function validateTileset(tileset, tilesetDirectory) {
 
 function validateTopLevel(tileset) {
     if (!defined(tileset.geometricError)) {
-        return 'Tileset must declare its geometricError as a top-level property.';
+        return errorMessage('Tileset must declare its geometricError as a top-level property.', undefined);
```
lilleyse: These at the top level don't need to go through `errorMessage`. Just returning the message is fine.

validator/lib/validateTileset.js
```diff
@@ -103,29 +104,29 @@ function validateTileHierarchy(root, tilesetDirectory) {
             var tileBox = tile.boundingVolume.box;
 
             if (defined(contentRegion) && defined(tileRegion) && !regionInsideRegion(contentRegion, tileRegion)) {
-                return 'content region [' + contentRegion + '] is not within tile region + [' + tileRegion + ']';
+                return errorMessage('content region [' + contentRegion + '] is not within tile region + [' + tileRegion + ']', tile.content);
```
lilleyse: For all of these just pass in `tile` instead of `tile.content`. The output message needs to show both the tile bounding volume and the content bounding volume.

validator/lib/validateTileset.js
```diff
@@ -173,3 +174,14 @@ function validateTileHierarchy(root, tilesetDirectory) {
             return message;
         });
 }
+
+function errorMessage(originalMessage, tile) {
+    if (defined(tile)) {
```
lilleyse: Related to the note above, it's fine for this helper function to always assume a `tile` is passed in.

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -46,7 +46,8 @@ describe('validateTileset', function() {
         delete tileset.root.children[0].geometricError;
         expect(validateTileset(tileset)
             .then(function(message) {
-                expect(message).toBe('Each tile must define geometricError');
+                var err = 'Each tile must define geometricError';
+                expect(message.slice(0, err.length)).toBe(err);
```
lilleyse: Try to avoid abbreviations, `error` is fine.

Rudraksha20:
@lilleyse Updated!

lilleyse:
Thanks @Rudraksha20!

PR #111, opened 2017-07-19
Added a composite of instances tileset.

tfili:


tfili:
cc @lilleyse

lilleyse:
I made some tweaks since the i3dm was embedding the gltf instead of referencing a url. It also saves out the glb file now.

tfili:
Oops. Should've caught that. Thanks @lilleyse.

PR #114, opened 2017-07-25
Using different texture

lilleyse:
The old texture somehow was not red-channel-only after being written to the gltf (color space conversion?). This texture works though and is a bit clearer that it's actually a texture instead of solid red.


shehzan10:
Yes certainly looks better than before.

Issue #110, opened 2017-07-13
Validator - check empty child within parent


lilleyse:
@Rudraksha20

From the roadmap:

> When a tile has no content (because it is just used for culling), its boundingVolume must be completely inside its parent tile's boundingVolume.

lilleyse:
This should also take a tile's `transform` (if it has one) into account. You can reference the code here for creating bounding volumes that are transformed: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/Cesium3DTile.js#L828-L873

PR #115, opened 2017-07-25
ValidateTileset checks for box in box bounding volume

Rudraksha20:
Issue #110

ValidateTileset checks for box in box bounding volume spatial coherence for tile with no content.

### DIFF
--- a/validator/lib/validateTileset.js
+++ b/validator/lib/validateTileset.js
@@ -97,39 +97,28 @@ function validateTileHierarchy(root, tilesetDirectory) {
         }
 
         if (defined(content) && defined(content.boundingVolume)) {
-            var contentBV = content.boundingVolume;
-            var tileBV = tile.boundingVolume;
-            var tileTransform = Matrix4.IDENTITY;
-            if(defined(tile.transform)) {
-                tileTransform = Matrix4.fromArray(tile.transform);
+            var outerTransform = Matrix4.IDENTITY;
+            if (defined(tile.transform)) {
+                outerTransform = Matrix4.fromArray(tile.transform);
             }
-            var contentTransform = Matrix4.IDENTITY;
-            if(defined(content.transform)) {
-                contentTransform = Matrix4.fromArray(content.transform);
-            }
-            var message = undefined;
-
-            message = checkBoundingVolume(contentBV, tileBV, contentTransform, tileTransform);
-            if(defined(message)) {
+            var innerTransform = Matrix4.IDENTITY;
+            var message = checkBoundingVolume(content.boundingVolume, tile.boundingVolume, innerTransform, outerTransform);
+            if (defined(message)) {
                 return 'content bounding volume is not within tile bounding volume: ' + message;
             }
         }
 
-        if (defined(parent) && !defined(content) && defined(tile.boundingVolume) && defined(parent.boundingVolume)) {
-            var tileBV = tile.boundingVolume;
-            var parentBV = parent.boundingVolume;
-            var tileTransform = Matrix4.IDENTITY;
-            if(defined(tile.transform)) {
-                tileTransform = Matrix4.fromArray(tile.transform);
+        if (defined(parent) && !defined(content)) {
+            var innerTransform = Matrix4.IDENTITY;
+            if (defined(tile.transform)) {
+                innerTransform = Matrix4.fromArray(tile.transform);
             }
-            var parentTransform = Matrix4.IDENTITY;
-            if(defined(parent.transform)) {
-                parentTransform = Matrix4.fromArray(parent.transform);
+            var outerTransform = Matrix4.IDENTITY;
+            if (defined(parent.transform)) {
+                outerTransform = Matrix4.fromArray(parent.transform);
             }
-            var message = undefined;
-
-            message = checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform);
-            if(defined(message)) {
+            var message = checkBoundingVolume(tile.boundingVolume, parent.boundingVolume, innerTransform, outerTransform);
+            if (defined(message)) {
                 return 'child bounding volume is not within parent bounding volume: ' + message;
             }
         }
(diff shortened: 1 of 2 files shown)


validator/lib/validateTileset.js
```diff
@@ -168,3 +191,38 @@ function validateTileHierarchy(root, tilesetDirectory) {
             return message;
         });
 }
+
+function checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform, type) {
```
lilleyse: This could be more generic so that it takes the bounding volumes instead of the boxes. Then the code here would basically include this code from above (plus code for handling the transforms) :

```
            var contentRegion = content.boundingVolume.region;
            var contentSphere = content.boundingVolume.sphere;
            var contentBox = content.boundingVolume.box;
            var tileRegion = tile.boundingVolume.region;
            var tileSphere = tile.boundingVolume.sphere;
            var tileBox = tile.boundingVolume.box;

            if (defined(contentRegion) && defined(tileRegion) && !regionInsideRegion(contentRegion, tileRegion)) {
                return 'content region [' + contentRegion + '] is not within tile region + [' + tileRegion + ']';
            }

            if (defined(contentSphere) && defined(tileSphere) && !sphereInsideSphere(contentSphere, tileSphere)) {
            }

            if (defined(contentBox) && defined(tileBox) && !boxInsideBox(contentBox, tileBox)) {
                return 'content box [' + contentBox + '] is not within tile box [' + tileBox + ']';
            }

            if (defined(contentBox) && defined(tileSphere) && !boxInsideSphere(contentBox, tileSphere)) {
                return 'content box [' + contentBox + '] is not within tile sphere [' + tileSphere + ']';
            }
```

validator/specs/lib/validateTilesetSpec.js
```diff
@@ -272,3 +330,26 @@ function createSampleTileset(tileBoundingVolume, contentBoundingVolume) {
     };
     return sampleTileset;
 }
+
+function createContentlessSampleTileset(parentBoundingVolume, childBoundingVolume, childTransform) {
```
lilleyse: A better name might be `createParentChildTileset`. It should also take in `parentTransform`.

validator/lib/validateTileset.js
```diff
@@ -12,7 +12,10 @@ var regionInsideRegion = utility.regionInsideRegion;
 var sphereInsideSphere = utility.sphereInsideSphere;
 var boxInsideBox = utility.boxInsideBox;
 var boxInsideSphere = utility.boxInsideSphere;
-
+var Matrix4 = Cesium.Matrix4;
+var Cartesian3 = Cesium.Cartesian3;
+var CesiumMath = Cesium.Math;
```
lilleyse: Sorry, duplicate.

validator/lib/validateTileset.js
Rudraksha20: So for this case you mean wherever we do those checks like 'regionInsideRegion' and such in the above code it should be replaced with 'checkBoundingVolume' and just take in the transforms and the bounding volumes and based on the type of the bounding volume operate on it?

Rudraksha20:
@lilleyse updated the code with the recommended changes.

Also I have made the function 'checkBoundingVolume' take in the bounding volume and inside it call the respective functions based on type provided and the stored bounding volume type. This will make the 'checkBoundingVolume' function more general.

Rudraksha20:
@lilleyse squashed some bugs that slipped by.

validator/lib/validateTileset.js
lilleyse: Roughly yes. However lets avoid sending in `BoxinBox` and other strings. `checkBoundingVolume` should know the bounding volume types by using the `defined` checks like in the code above.

I think the code right now is constrained by how the error messages are created. `checkBoundingVolume` could return a generic message like `'box [' + box + '] is not within sphere [' + sphere + ']'`. Then since `checkBoundingVolume` is called by different areas (child vs parent and content vs tile), those can add onto the error message that is returned. Something like: `'content bounding volume is not within tile bounding volume:' + checkBoundingVolumeErrorMessage`.

With a few structural changes like this I think this area can become a lot cleaner.

lilleyse:
Watch out for merge conflicts - sphere inside box was just added: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/113

Rudraksha20:
@lilleyse Updated and resolved merge conflicts!

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
-            var contentSphere = content.boundingVolume.sphere;
-            var contentBox = content.boundingVolume.box;
-            var tileRegion = tile.boundingVolume.region;
-            var tileSphere = tile.boundingVolume.sphere;
-            var tileBox = tile.boundingVolume.box;
-
-            if (defined(contentRegion) && defined(tileRegion) && !regionInsideRegion(contentRegion, tileRegion)) {
-                return 'content region [' + contentRegion + '] is not within tile region + [' + tileRegion + ']';
+            var contentBV = content.boundingVolume;
+            var tileBV = tile.boundingVolume;
```
lilleyse: We try to avoid abbreviations, so go with `contentBoundingVolume` and `tileBoundingVolume` throughout.

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            var tileTransform = Matrix4.IDENTITY;
+            if(defined(tile.transform)) {
```
lilleyse: Just a small style thing, we usually put spaces after the `if`:

`if (defined(tile.transform)) {`

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+                tileTransform = Matrix4.fromArray(tile.transform);
             }
-
-            if (defined(contentSphere) && defined(tileSphere) && !sphereInsideSphere(contentSphere, tileSphere)) {
-                return 'content sphere [' + contentSphere + '] is not within tile sphere + [' + tileSphere + ']';
+            var contentTransform = Matrix4.IDENTITY;
```
lilleyse: `content` does not have a transform, so you can just pass `tileTransform` twice to `checkBoundingVolume`, or just the identity in this case.

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            if(defined(content.transform)) {
+                contentTransform = Matrix4.fromArray(content.transform);
             }
+            var message = undefined;
 
-            if (defined(contentBox) && defined(tileBox) && !boxInsideBox(contentBox, tileBox)) {
-                return 'content box [' + contentBox + '] is not within tile box [' + tileBox + ']';
+            message = checkBoundingVolume(contentBV, tileBV, contentTransform, tileTransform);
```
lilleyse:
```
var message = undefined;

message = checkBoundingVolume(contentBV, tileBV, contentTransform, tileTransform);
```

Can be `var message = checkBoundingVolume(contentBV, tileBV, contentTransform, tileTransform);`

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            if(defined(message)) {
```
lilleyse: Same style comment here.

validator/lib/validateTileset.js
lilleyse: Though in this case it may not be necessary to store these as variables.

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+                return 'content bounding volume is not within tile bounding volume: ' + message;
             }
+        }
 
-            if (defined(contentBox) && defined(tileSphere) && !boxInsideSphere(contentBox, tileSphere)) {
-                return 'content box [' + contentBox + '] is not within tile sphere [' + tileSphere + ']';
+        if (defined(parent) && !defined(content) && defined(tile.boundingVolume) && defined(parent.boundingVolume)) {
```
lilleyse: `defined(tile.boundingVolume) && defined(parent.boundingVolume)` - these should always be defined so no need to check here. However we may need to add a fix in another PR to check that these exist.

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,40 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            if(defined(parent.transform)) {
+                parentTransform = Matrix4.fromArray(parent.transform);
+            }
+            var message = undefined;
 
-            if (defined(contentSphere) && defined(tileBox) && !sphereInsideBox(contentSphere, tileBox)) {
-                return 'content sphere [' + contentSphere + '] is not within tile box [' + tileBox + ']';
+            message = checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform);
+            if(defined(message)) {
+                return 'child bounding volume is not within parent bounding volume: ' + message;
```
lilleyse: In general same comments from above apply here too.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
             return message;
         });
 }
+
+function checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform) {
+    var returnString = undefined;
+
+    if (defined(tileBV) && defined(parentBV) && (defined(tileBV.box) || defined(tileBV.sphere) || defined(tileBV.region)) && (defined(parentBV.box) || defined(parentBV.sphere) || defined(parentBV.region))) {
```
lilleyse: Similar to above this check will not be needed, but a separate PR will check for the existence of bounding volumes.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
             return message;
         });
 }
+
+function checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform) {
```
lilleyse: Perhaps better terminology is `innerBoundingVolume`, `outerBoundingVolume`, `innerTransform`, and `outerTransform`, just a bit more generic and matches the naming in `utility.js`.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+    var returnString = undefined;
+
+    if (defined(tileBV) && defined(parentBV) && (defined(tileBV.box) || defined(tileBV.sphere) || defined(tileBV.region)) && (defined(parentBV.box) || defined(parentBV.sphere) || defined(parentBV.region))) {
+        if (defined(tileBV.box) && defined(parentBV.box)) {
+            // Box in Box check
+            var transformed_tileBox = getTransformedBox(tileBV.box, tileTransform);
+            var transformed_parentBox = getTransformedBox(parentBV.box, parentTransform);
+            if (boxInsideBox(transformed_tileBox, transformed_parentBox) !== true) {
```
lilleyse: ` if (!boxInsideBox(transformed_tileBox, transformed_parentBox)) {` is cleaner.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
         });
 }
+
+function checkBoundingVolume(tileBV, parentBV, tileTransform, parentTransform) {
+    var returnString = undefined;
+
+    if (defined(tileBV) && defined(parentBV) && (defined(tileBV.box) || defined(tileBV.sphere) || defined(tileBV.region)) && (defined(parentBV.box) || defined(parentBV.sphere) || defined(parentBV.region))) {
+        if (defined(tileBV.box) && defined(parentBV.box)) {
+            // Box in Box check
+            var transformed_tileBox = getTransformedBox(tileBV.box, tileTransform);
```
lilleyse: Use camel case for these and throughout.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+            var transformed_parentBox = getTransformedBox(parentBV.box, parentTransform);
+            if (boxInsideBox(transformed_tileBox, transformed_parentBox) !== true) {
+                returnString = 'box [' + tileBV.box + '] is not within box [' + parentBV.box + ']';
+                return returnString;
+            } else {
+                return returnString;
+            }
```
lilleyse: All these returns could be simplified. Just set the `returnString`, and at the end of `checkBoundingVolume` return `returnString`. Also `message` is a better name than `returnString`.

validator/lib/validateTileset.js
```diff
@@ -173,3 +184,103 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+        }
+    } else {
+        return returnString;
+    }
+}
+
+function getTransformedBox(box, transform) {
+    var scratchMatrix = new Matrix3();
+    var scratchHalfAxes = new Matrix3();
+    var scratchCenter = new Cartesian3();
```
lilleyse: Put scratches outside of the function. That way they are not allocation for every function call, but instead exist in global scope.

validator/lib/validateTileset.js
Rudraksha20: sorry about that must have missed it.

Rudraksha20:
@lilleyse updated!

validator/lib/validateTileset.js
```diff
@@ -95,31 +97,29 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
-                return 'content region [' + contentRegion + '] is not within tile region + [' + tileRegion + ']';
+            var outerTransform = Matrix4.IDENTITY;
+            if (defined(tile.transform)) {
+                outerTransform = Matrix4.fromArray(tile.transform);
             }
-
-            if (defined(contentSphere) && defined(tileSphere) && !sphereInsideSphere(contentSphere, tileSphere)) {
-                return 'content sphere [' + contentSphere + '] is not within tile sphere + [' + tileSphere + ']';
+            var innerTransform = Matrix4.IDENTITY;
+            var message = checkBoundingVolume(content.boundingVolume, tile.boundingVolume, innerTransform, outerTransform);
```
lilleyse: I'm getting some warnings since `outerTransform`, `innerTransform` and `message` are declared multiple times in the same function scope. Can you declare these above, like:

```
        var outerTransform;
        var innerTransform
        var mesage;

        if (defined(content) && defined(content.boundingVolume)) {
            outerTransform = Matrix4.IDENTITY;
            if (defined(tile.transform)) {
                outerTransform = Matrix4.fromArray(tile.transform);
            }
            innerTransform = Matrix4.IDENTITY;
            message = checkBoundingVolume(content.boundingVolume, tile.boundingVolume, innerTransform, outerTransform);
            if (defined(message)) {
                return 'content bounding volume is not within tile bounding volume: ' + message;
            }
        }
```

validator/lib/validateTileset.js
```diff
@@ -173,3 +173,92 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+
+function checkBoundingVolume(innerBoundingVolume, outerBoundingVolume, innerTransform, outerTransform) {
+    var message;
+    var transformedInnerTile;
+    var transformedOuterTile;
+
+    if (defined(innerBoundingVolume.box) && defined(outerBoundingVolume.box)) {
+        // Box in Box check
+        var transformedInnerTile = getTransformedBox(innerBoundingVolume.box, innerTransform);
+        var transformedOuterTile = getTransformedBox(outerBoundingVolume.box, outerTransform);
```
lilleyse: Just remove the `var` in front of these since they are already declared above.

validator/lib/validateTileset.js
```diff
@@ -173,3 +173,92 @@ function validateTileHierarchy(root, tilesetDirectory) {
...
+        if (!sphereInsideBox(transformedInnerTile, transformedOuterTile)) {
+            message = 'sphere [' + innerBoundingVolume.sphere + '] is not within box [' + outerBoundingVolume.box + ']';
+            return message;
+        }
+        return message;
+    } else {
+        // Add more test cases here!
+        return message;
+    }
```
lilleyse: Remove this last `else` block.

Rudraksha20:
@lilleyse Updated!

lilleyse:
Thanks! I fixed some merge conflicts related to the tests - looks good!

PR #121, opened 2017-08-10
Validator - check that RGBA/RGB/RGB565/CONSTANT_RGBA are in range

### LINKED ISSUE #119: Validator - check that RGBA/RGB/CONSTANT_RGBA are in range
@rms13

Check that `RGBA`/`RGB`/`CONSTANT_RGBA` are in range. The values should be 0-255 inclusive.

This one will be similar to the normal checking.

rms13:
fixes #119 

Checks that the values of RGBA (array), RGB (array) and CONSTANT_RGBA are between 0-255, and that the values of RGB565 (array) are between 0-65535.

### DIFF
--- a/validator/lib/validatePnts.js
+++ b/validator/lib/validatePnts.js
@@ -221,7 +221,6 @@ function validatePnts(content) {
         featureTable.featuresLength = pointsLength;
         componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.RGB565.componentType, 'UNSIGNED_INT', 1)); // UNSIGNED_BYTE
         var rgb565 = featureTable.getPropertyArray('RGB565', componentDatatype, 1);
-        //console.log(rgb565);
         var max = Math.max(...rgb565);
         var min = Math.min(...rgb565);
         if (min < 0 || max > 65535) {
--- a/validator/lib/validatePnts.js
+++ b/validator/lib/validatePnts.js
@@ -185,14 +185,13 @@ function validatePnts(content) {
 
     var featureTable = new Cesium3DTileFeatureTable(featureTableJson, featureTableBinary);
     featureTable.featuresLength = pointsLength;
-    var i;
     var componentDatatype;
 
     if (defined(featureTableJson.BATCH_ID)) {
         componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.BATCH_ID.componentType, 'UNSIGNED_SHORT'));
         var batchIds = featureTable.getPropertyArray('BATCH_ID', componentDatatype, 1);
         var length = batchIds.length;
-        for (i = 0; i < length; i++) {
+        for (var i = 0; i < length; i++) {
             if (batchIds[i] >= featureTableJson.BATCH_LENGTH) {
                 return 'All the BATCH_IDs must have values less than feature table property BATCH_LENGTH.';
             }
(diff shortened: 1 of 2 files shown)


validator/lib/validatePnts.js
```diff
@@ -196,6 +198,42 @@ function validatePnts(content) {
         }
     }
 
+    if (defined(featureTableJson.RGBA)) {
+        featureTable.featuresLength = pointsLength * 4;
+        componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.RGBA.componentType, 'UNSIGNED_SHORT', 4));
+        var rgba = featureTable.getPropertyArray('RGBA', componentDatatype, 1);
```
lilleyse: Same note as https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/116/files#r133109885

validator/specs/lib/validatePntsSpec.js
```diff
@@ -257,4 +257,169 @@ describe('validate pnts', function() {
         });
         expect(validatePnts(pnts)).toBeUndefined();
     });
+
+    it('returns error message when RGBA is out of range [Test using feature table JSON]', function() {
```
lilleyse: Can you also add a test for when it succeeds?

validator/lib/validatePnts.js
```diff
@@ -196,6 +198,42 @@ function validatePnts(content) {
         }
     }
 
+    if (defined(featureTableJson.RGBA)) {
+        featureTable.featuresLength = pointsLength * 4;
+        componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.RGBA.componentType, 'UNSIGNED_SHORT', 4));
```
lilleyse: `UNSIGNED_SHORT` should be `UNISNGED_BYTE`.

lilleyse: Remove the `, 4` at the end.

validator/lib/validatePnts.js
```diff
@@ -196,6 +198,42 @@ function validatePnts(content) {
...
+        var rgba = featureTable.getPropertyArray('RGBA', componentDatatype, 1);
+        var max = Math.max(...rgba);
+        var min = Math.min(...rgba);
+        if (min < 0 || max > 255) {
+            return 'values of RGBA must be in the range 0-255 inclusive';
+        }
+    } else if (defined(featureTableJson.RGB)) {
+        featureTable.featuresLength = pointsLength * 3;
+        componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.RGB.componentType, 'UNSIGNED_SHORT', 3));
```
lilleyse: Should be `UNSIGNED_BYTE`.

validator/lib/validatePnts.js
```diff
@@ -196,6 +198,42 @@ function validatePnts(content) {
...
+        var rgb = featureTable.getPropertyArray('RGB', componentDatatype, 1);
+        var max = Math.max(...rgb);
+        var min = Math.min(...rgb);
+        if (min < 0 || max > 255) {
+            return 'values of RGB must be in the range 0-255 inclusive';
+        }
+    } else if (defined(featureTableJson.RGB565)) {
+        featureTable.featuresLength = pointsLength;
+        componentDatatype = ComponentDatatype.fromName(defaultValue(featureTableJson.RGB565.componentType, 'UNSIGNED_INT', 1));
```
lilleyse: Should be `UNSIGNED_SHORT`.

validator/lib/validatePnts.js
rms13: I have purposely set the parameter to `UNSIGNED_SHORT`. When I set it to `UNSIGNED_BYTE`, the values, if out of range, will be clipped when creating an array using `featureTable.getPropertyArray()`. So I am using higher precision type to catch invalid values. Same for the other checks.

lilleyse: We may have talked about this offline before.... but here are my latest thoughts: using a higher precision is unsafe because it distorts the values that are retrieved from the feature table binary and may result in out-of-bounds issues. Ultimately these range checks for colors are not super useful for binary properties, but are helpful for array-based properties.

rms13:
@lilleyse Should be good now.

PR #125, opened 2017-08-25
Validate glTF BatchID are in range

Rudraksha20:
Work in Progress.


Rudraksha20:
@lilleyse 

- [done] Test for validating BatchID are not greater than (Batch_Length -1).

PR #135, opened 2018-01-05
Vertex color tileset

lilleyse:
Generates a tileset that uses vertex colors. CC https://github.com/AnalyticalGraphicsInc/cesium/pull/6089

### DIFF
--- a/samples-generator/lib/Mesh.js
+++ b/samples-generator/lib/Mesh.js
@@ -16,8 +16,7 @@ var sizeOfUint16 = 2;
 var sizeOfFloat32 = 4;
 
 var whiteOpaqueMaterial = new Material({
-    diffuse : [1.0, 1.0, 1.0, 1.0],
-    ambient : [0.2, 0.2, 0.2, 1.0]
+    diffuse : [1.0, 1.0, 1.0, 1.0]
 });
 
 /**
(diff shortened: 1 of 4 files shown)


pjcozzi:
@emackey any chance you can look at this?

emackey:
I haven't had a chance to dig into this repo and its uses yet, but the code looks OK at a glance.

pjcozzi:
Thanks @emackey.  @shehzan10 could you take a closer look?

samples-generator/lib/Mesh.js
```diff
@@ -130,6 +136,23 @@ Mesh.prototype.setPositionsRelativeToCenter = function() {
...
+/**
+ * Bake materials as vertex colors. Use the default white opaque material.
+ */
+Mesh.prototype.transferMaterialToVertexColors = function() {
+    var material = this.material;
+    this.material = whiteOpaqueMaterial;
+    var vertexCount = this.getVertexCount();
+    var vertexColors = new Array(vertexCount * 4);
+    this.vertexColors = vertexColors;
+    for (var i = 0; i < vertexCount; ++i) {
```
shehzan10: What do you think about using different per-vertex colors to give it a "colorful" feel? I feel like that would make a better sample.

lilleyse: That would demonstrate vertex colors better, but I think this is alright for now. It's also easier to work with for the color blend mode test in Cesium which requires the center building to be red.

PR #136, opened 2018-01-17
Create point cloud tileset

lilleyse:
Adds a point clouds tileset that contains more than 1 tile. For https://github.com/AnalyticalGraphicsInc/cesium/pull/6113

### DIFF
--- a/samples-generator/bin/3d-tiles-samples-generator.js
+++ b/samples-generator/bin/3d-tiles-samples-generator.js
@@ -2224,7 +2224,7 @@ function createTilesetPoints() {
     var tilesetPath = path.join(tilesetDirectory, 'tileset.json');
 
     var pointsLength = 1000;
-    var parentTileWidth = 20.0;
+    var parentTileWidth = 10.0;
     var parentTileHalfWidth = parentTileWidth / 2.0;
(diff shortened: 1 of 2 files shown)


shehzan10:
The point cloud seems very sparse up close and much better from farther away. We could make it more dense.

lilleyse:
I shrunk the size by half, but I'm not sure if it helps that much when using the zoom-to code:

```
tileset.readyPromise.then(function() {
    var boundingSphere = tileset.boundingSphere;
    viewer.camera.viewBoundingSphere(boundingSphere, new Cesium.HeadingPitchRange(0.0, -0.5, boundingSphere.radius / 4.0));
    viewer.camera.lookAtTransform(Cesium.Matrix4.IDENTITY);
}).otherwise(function(error) {
    throw(error);
});
```

I would like to keep the POINTS_LENGTH at 1000 so as not to bloat Cesium too much. The other point clouds have the same problem ultimately.

shehzan10:
> I shrunk the size by half, but I'm not sure if it helps that much when using the zoom-to code:

Yeah, I realized this too. I think it really depends on how the camera is placed when viewing the tileset.

Looks good otherwise.

PR #139, opened 2018-03-21
glTF 2.0 in samples generator

lilleyse:
Updated the samples generator project to use glTF 2.0.

The tilesets all load in Cesium and the embedded glTFs pass [glTF Validator](https://github.com/KhronosGroup/glTF-Validator).

Not quite ready for merging yet because
* [x] Some samples are commented out. This includes anything that used features from `gltf-pipeline#1.0` including texture compression, quantization, or oct encoding.
    * [x] Tree billboard tileset requires [KHR_techniques_webgl](https://github.com/KhronosGroup/glTF/issues/1217) to be ready
* [x] package.json uses gltf-pipeline 1.0.5 but requires gltf-pipeline 2.0.0 which isn't released yet.

After merge:
* [ ] Update [3d-tiles-samples](https://github.com/AnalyticalGraphicsInc/3d-tiles-samples) after [KHR_techniques_webgl in Cesium](https://github.com/AnalyticalGraphicsInc/cesium/pull/6805) is merged. The tree billboard sample requires that extension.

PR #140, opened 2018-03-29
Fix geometric error in TilesetOfTilesets example

lilleyse:
Problems in TilesetOfTilesets that were exposed by https://github.com/AnalyticalGraphicsInc/cesium/pull/6390


lilleyse:
Moved into #139

PR #144, opened 2018-04-10
upgrade: support relative url in tile file

huiyiqun:
This patch to solve the problem that when there is relative url in tile file, the upgrade procedure will fail with following message:

```
glTF model references external files but no basePath is supplied
```

which is raised from [here](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/f58a3a87234affe2f1c5961ed894ed3c310fd992/lib/loadGltfUris.js#L100)

### DIFF
--- a/tools/lib/optimizeGlb.js
+++ b/tools/lib/optimizeGlb.js
@@ -39,7 +39,7 @@ function optimizeGlb(glbBuffer, options) {
         }
     }
     fixBatchIdSemantic(gltf);
-    return loadGltfUris(gltf)
+    return loadGltfUris(gltf, options)
         .then(function() {
             return Pipeline.processJSONWithExtras(gltf, options)
                 .then(function(gltf) {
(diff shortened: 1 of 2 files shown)


lilleyse:
Thanks @huiyiqun! The code looks good to me.

If you haven't already, please send in a [CLA](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla) so that we can merge your changes.

huiyiqun:
@lilleyse Well, the CLA has just been sent.

huiyiqun:
@lilleyse Hi, any progress here?

pjcozzi:
Thanks again, @huiyiqun, we received your CLA.

It may take a few days for someone to review this as the team has been in training.

lilleyse:
@huiyiqun sorry about the delay, but it all looks good. I confirmed that the fix works.

PR #153, opened 2018-08-29
Add extras dictionary to tileset

OmarShehata:
This PR adds an extras dictionary to the tileset as well as to one of the children. This is to test adding support for extras in CesiumJS (AnalyticalGraphicsInc/cesium#6974).

I branched off of `2.0-samples-generator` to get the updates in that PR (https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/139).


lilleyse:
Thanks @OmarShehata

PR #170, opened 2018-12-20
avoid calling getJsonBufferPadded twice

wassname:
This avoids calling getJsonBufferPadded on the featuretablejson twice. 

Once in readGlbWriteI3dm and a second time in glbToI3dm which would result in something like. 

```
{"type":"Buffer","data":[123,34,73,78,83,84,65,78,67,69,83,95,76,69,78,71,84,72,34,58,49,44,34,80,79,83,73,84,73,79,78,34,58,123,34,98,121,116,101,79,102,102,115,101,116,34,58,48,125,125,32,32,32,32,32,32]}
```
That would in turn lead to a cesium loading error ""Error: Feature table global property: INSTANCES_LENGTH must be defined""

### DIFF
--- a/tools/bin/3d-tiles-tools.js
+++ b/tools/bin/3d-tiles-tools.js
@@ -319,10 +319,9 @@ function readGlbWriteI3dm(inputPath, outputPath, force) {
                             byteOffset : 0
                         }
                     };
-                    var featureTableJsonBuffer = getJsonBufferPadded(featureTable);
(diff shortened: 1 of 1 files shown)


lilleyse:
Hi @wassname, thanks for the fix. Can you fill out a CLA so we can merge this?

https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

PR #172, opened 2019-03-14
Fix Node 10

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/161

One test is failing locally in Linux:

```
1) optimizeGlb compresses textures in a glb using the gltf-pipeline
  Message:
    Failed: Expected promise to resolveError: spawn /home/slilley/Code/3d-tiles-tools/tools/node_modules/gltf-pipeline/bin/linux/crunch EACCES
```

because the gltf-pipeline texture compression binaries don't have their executable bit set after package is install. I disabled the test for now, but more reason to eventually wrap up [2.0-tools](https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/tree/2.0-tools) if time allows.

### DIFF
--- a/tools/package.json
+++ b/tools/package.json
@@ -22,27 +22,27 @@
     "node": ">=4.0.0"
   },
   "dependencies": {
-    "bluebird": "^3.5.1",
-    "cesium": "^1.39",
-    "fs-extra": "^4.0.2",
-    "gltf-pipeline": "^1.0.2",
-    "klaw": "^2.1.0",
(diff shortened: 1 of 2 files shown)


shehzan10:
There are eslint errors for `var`/`let`/`const`. I assume we can fix those using eslint itself, do we want to do that in this PR?

Otherwise looks good. Tested on Windows and Linux.

lilleyse:
Let's fix that later.

PR #180, opened 2020-01-13
Add glTF 2.0 validation

erikdahlstrom:
Should be considered a draft, but seems to work for the cases I've tried.

Using the js version of glTF-Validator avoids the requirement to have Dart SDK installed, all that's needed is 'npm install'.


lilleyse:
From a quick review, looks like all good changes. I'm fine with merging even if it's a draft.

lilleyse:
Thanks @erikdahlstrom

PR #181, opened 2020-02-10
B3DM featuretable may contain 'RTC_CENTER'.

erikdahlstrom:
Validation failed if feature table contained 'RTC_CENTER'.


lilleyse:
Thanks @erikdahlstrom!

PR #185, opened 2020-02-20
Move the negation to the right place.

erikdahlstrom:


lilleyse:
Thanks @erikdahlstrom!

PR #186, opened 2020-02-25
B3dm to cesium 3dtiles batch table

Samulus:
~~wip, not ready to merge yet~~
needs reviews

Has:
* Full GLB support
* Full GLTF support
* Works with non binary batchtable properties directly embedded in the extension section
* Works with binary batchtable properties (creates additional accessors / bufferviews / an additional buffer with the `code` / `cartographic` properties concatenated together
* ~~FeatureTables~~ (Deprecated, RTC_CENTER is becoming a GLTF node, will make a demo using this in a follow up PR)
* ~~Avoid writing redundant `id` property~~ see comment in PR about this
* Unittests

Needs:

* Multiple batch table support (currently only works with one batch table, this is probably going in a follow up PR)


samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,116 @@
...
+
+    if (defined(humanAttributes)) {
+        var batchLength = humanAttributeNames.length > 0 ? humanAttributes[humanAttributeNames[0]].length : 0;
+        var newBatchTable = {batchLength: batchLength, properties: {}};
+
+        for (i = 0; i < humanAttributeNames.length; ++i) {
+            var values = humanAttributes[humanAttributeNames[i]];
+            // TODO: Should we hardcode a if batchAttributes[i] === 'id' here?
+            //       It could fail if we have an batch attribute named id that doesn't use
+            //       an implicit indexing structure (i.e 2, 4, 8, 1, 9, 30...);
```
lilleyse: I think it's working fine, the id property in the glTF output looked correct to me.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -91,24 +95,54 @@ function createBuildingsTile(options) {
...
+        batchTableJson = combine(batchTableJson, batchTableJsonAndBinary.json);
         return {
-            b3dm : b3dm,
+            b3dm : createB3dm(b3dmOptions),
             batchTableJson : batchTableJson
         };
+    }).catch(function(e) {
+        console.log(e.message);
+        console.log(e.stack);
     });
```
lilleyse: I think these errors should get propagated automatically

samples-generator/lib/createBuildingsTile.js
```diff
@@ -91,24 +95,54 @@ function createBuildingsTile(options) {
...
-            glb : glb,
+        deprecated : deprecated1 || deprecated2,
+        useGlb: useGlb,
+        useGltf: useGltf,
+        deferGlbConversion: useGlb || useGltf // TODO: This is ugly, but allows us to refactor this
+                                              //       without breaking other functions that potentially always expect
+                                              //       createGltf to return a promise that resolves to a glb.
+                                              //       Instead, it returns a promise that resolves to a gltf
+                                              //       hence 'glbOrGltf'. We can do further post-processing on the gltf,
+                                              //       and convert it to a GLB when necessary later.
```
lilleyse: Maybe there should be a `createGltf` and `createGlb` that take the same options. `createGlb` could just call `createGltf` and do the conversion. Rather than listing all the options twice `createGlb` doc could look like https://github.com/CesiumGS/gltf-pipeline/blob/master/lib/gltfToGlb.js#L11-L17.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -161,13 +195,20 @@ function generateBatchTableBinary(buildings) {
 
     var batchTableJson = {
         cartographic : {
+            name : 'CartographicBuffer',
             byteOffset : 0,
-            componentType : 'DOUBLE',
-            type : 'VEC3'
+            byteLength : cartographicBuffer.length,
+            componentType : 0x1406, // TODO: Logical error, not sure what to do, 'DOUBLE' was here originally but GLTF only supports Floats
```
lilleyse: We'll have to use floats for now. I noticed this too in https://github.com/CesiumGS/glTF/pull/1

samples-generator/lib/createBuildingsTile.js
```diff
@@ -161,13 +195,20 @@ function generateBatchTableBinary(buildings) {
...
+
         code : {
+            name: 'Code',
             byteOffset : cartographicBuffer.length,
-            componentType : 'UNSIGNED_BYTE',
+            count: codeBuffer.length,
+            byteLength: codeBuffer.length,
+            componentType : 0x1401,
             type : 'SCALAR'
         }
```
lilleyse: These changes might go against the "legacy" batch table binary schema. `byteLength` and `count` are not valid properties in https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/batchTable.schema.json#L19-L40.

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -196,6 +202,13 @@ var ulTileOptions = {
 };
 
 var promises = [
+    // CESIUM_3dtiles_batch_table
+    createBatchedWithBatchTableGltf(),
+    createBatchedWithBatchTableBinaryGltf(),
```
lilleyse: Are the binary properties getting added to the glTF? I just see `id`, `Longitude`, `Latitude`, and `Height`.

samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,116 @@
...
+function sortByByteOffset(a, b) {
+    if (a.byteOffset < b.byteOffset ){
+        return -1;
+    }
+
+    if (a.byteOffset < b.byteOffset) {
+        return 1;
+    }
+
+    return 0;
```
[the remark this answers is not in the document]
Samulus: https://flak.tedunangst.com/post/subtraction-is-not-comparison // this was the rationale for not using the simplified version, I'll change it though if we're not worried about overflow underflow issues.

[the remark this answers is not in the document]
lilleyse: Ah thanks for that out. I don't think we're worried about overflow here but let's keep your code since it's the correct thing to do.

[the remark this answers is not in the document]
Samulus: It is bugged though, the second line should say `>`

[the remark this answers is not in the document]
lilleyse: However just make sure to fix the code. `a.byteOffset < b.byteOffset` is checked twice.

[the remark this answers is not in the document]
lilleyse: Ah you beat me to it. https://github.com/CesiumGS/3d-tiles-validator/pull/186#discussion_r385234792

samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,116 @@
...
+        throw new Error('humanAttributes or binaryAttributes must be defined');
+    }
+
+    if (defined(binaryAttributes) && !defined(sharedBinaryBuffer)) {
+        throw new Error('sharedBinaryBuffer must be defined if binaryAttributes is defined');
+    }
+
+    if (!defined(gltf['extensionsUsed'])) {
+        gltf['extensionsUsed'] = [];
+    }
```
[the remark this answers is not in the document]
Samulus: That function isn't exported, so we'll keep the manual extension code

samples-generator/bin/3d-tiles-samples-generator.js
Samulus: They are now

samples-generator/lib/createBuildingsTile.js
Samulus: added a legacy mode when !glb and !gltf that uses the old behavior selectively :+1:

samples-generator/lib/createBuildingsTile.js
Samulus: they're split now

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -196,6 +202,13 @@ var ulTileOptions = {
...
+
+    createBatchedWithBatchTableGlb(),
+    createBatchedWithBatchTableBinaryGlb(),
+
```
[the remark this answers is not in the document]
Samulus: added yargs / mutual exclusion check with `--legacy`, `--glb` and `--gltf` as options

samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param {Buffer} sharedBinaryBuffer Currently the samples generator only supports creating
+ *                                    ONE additional Buffer with all the binary attributes inside of it.
+ *                                    Each value in `binaryAttributes` must correspond to this buffer. It
+ *                                    is an error to call this function using binaryAttributes without also
+ *                                    providing this value.
+ */
+
+function create3dtilesBatchTableExt(gltf, humanAttributes, binaryAttributes, sharedBinaryBuffer) {
+    gltf['extensionsUsed'] = [cesium3dTilesBatch];
```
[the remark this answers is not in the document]
Samulus: Can't use that function because it's not exported, but fixed the overwriting logic.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -189,3 +226,46 @@ function generateBatchTableBinary(buildings) {
...
+    var buildingsLength = buildings.length;
+    var cartographicBuffer = Buffer.alloc(buildingsLength * 3 * sizeOfDouble);
+    var codeBuffer = Buffer.alloc(buildingsLength * sizeOfUint8);
+
+    var batchTableJson = {
+        cartographic : {
+            name : 'CartographicBuffer',
+            byteOffset : 0,
+            byteLength : cartographicBuffer.length,
+            componentType : 0x1406, // TODO: Logical error, not sure what to do, 'DOUBLE' was here originally but GLTF only supports Floats
```
lilleyse: Let's remove the TODO for now and go with 32-bit floats until we have a better solution. Make sure that `cartographicBuffer` uses `sizeOfFloat` instead of `sizeOfDouble` above and uses `writeFloatLE` instead of `writeDoubleLE` below.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -189,3 +226,46 @@ function generateBatchTableBinary(buildings) {
...
+            type : 'VEC3',
+            count : buildingsLength
+        },
+
+        code : {
+            name: 'Code',
+            byteOffset : cartographicBuffer.length,
+            count: codeBuffer.length,
+            byteLength: codeBuffer.length,
+            componentType : 0x1401,
```
lilleyse:
```suggestion
            componentType : 0x1401, // UNSIGNED_BYTE
```

samples-generator/lib/createBuildingsTile.js
lilleyse:
```suggestion
            componentType : 0x1406, // FLOAT
```

samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {batchTableAttribute} batchTableAttributes List of batch table attributes. The function will intelligently try to detect
+ *                                                   if the attribute is a human readable value (a key value where the value is
+ *                                                   just an array of non-binary data) or if the values are more keys describing the
+ *                                                   layout of binary data in a binary buffer.
+ * @param {Buffer}              sharedBinaryBuffer   Binary buffer. Must be defined if using at least 1 binary batch attribute.
+ *                                                   This function currently assumes that all of the binary batch attributes in
+ *                                                   batchTableAttributes are directly referring to this buffer.
+ */
+
+function create3dtilesBatchTableExt(gltf, batchTableAttributes, sharedBinaryBuffer) {
```
lilleyse: Maybe just `createBatchTableExtension`?

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -926,24 +942,60 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
...
+                return Promise.all([
+                    saveJson(tilesetPath, createTilesetJsonSingle(tilesetOptions), prettyJson),
+                    saveJson(tilePath, result, prettyJson)
+                ]);
+            }
+
+            // old .b3dm
             var b3dm = result.b3dm;
             var batchTableJson = result.batchTableJson;
             tilesetOptions.properties = getProperties(batchTableJson);
```
lilleyse: The new gltf/glb code paths don't return the batch table json which is needed to generate the `properties` object in `tileset.json`.

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -926,24 +942,60 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
...
+            if (argv.glb) {
+                return Promise.all([
+                    fsExtra.outputFile(tilePath, result),
+                    saveJson(tilesetPath, createTilesetJsonSingle(tilesetOptions), prettyJson)
+                ]);
+            }
+
+            else if (argv.gltf) {
+                return Promise.all([
+                    saveJson(tilesetPath, createTilesetJsonSingle(tilesetOptions), prettyJson),
```
lilleyse: In non-legacy mode let's bump tileset.json `asset.version` to 1.1

```
  "asset": {
    "version": "1.1"
  },
```

samples-generator/lib/createBuildingsTile.js
```diff
@@ -91,24 +104,47 @@ function createBuildingsTile(options) {
         featureTableJson.RTC_CENTER = Cartesian3.pack(batchedMesh.getCenter(), new Array(3));
```
lilleyse: `RTC_CENTER` is getting lost in the gltf/glb code path. If `--gltf` or `--glb` is set `createGltf` should create a node with a translation to replace the old `RTC_CENTER`/feature table behavior.

samples-generator/lib/createGltf.js
```diff
@@ -1,18 +1,11 @@
 'use strict';
```
lilleyse: The gltf/glb code path should be naming the batch id vertex attribute `_BATCHID_0`, not `_BATCHID`

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
```
lilleyse: If running in non-legacy mode don't write samples for `BatchedDeprecated1` and `BatchedDeprecated2`. This is way legacy that the 3D Tiles Next doesn't have to worry about.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -189,3 +225,46 @@ function generateBatchTableBinary(buildings) {
...
 }
+
+function generateBatchTableBinary(buildings) {
+    var buildingsLength = buildings.length;
+    var cartographicBuffer = Buffer.alloc(buildingsLength * 3 * sizeOfDouble);
+    var codeBuffer = Buffer.alloc(buildingsLength * sizeOfUint8);
+
+    var batchTableJson = {
+        cartographic : {
+            name : 'CartographicBuffer',
```
lilleyse: For consistency with the legacy batch table names rename these properties:
* `CartographicBuffer` -> `cartographic`
* `Code` -> `code`

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -952,12 +1004,12 @@ function saveBatchedTileset(tilesetName, tileOptions, tilesetOptions) {
                 var dataUri = new DataUri();
                 dataUri.format('.b3dm', b3dm);
```
lilleyse: `BatchedWithContentDataUri` is not getting written as a data uri in non-legacy mode

samples-generator/specs/bin/CESIUM_3dtiles_batch_table_Spec.js
```diff
@@ -0,0 +1,108 @@
+'use strict';
+var minimalTriangleGLTF = require('./minimalTriangleGLTF');
+var create3dtilesBatchTableExt = require ('../../lib/create3dtilesBatchTableExt');
+
+describe('CESIUM_3dtiles_batch_table.js', function() {
```
lilleyse: Style: drop the `.js`. Also use the name of the file instead of `CESIUM_3dtiles_batch_table`

samples-generator/specs/bin/CESIUM_3dtiles_batch_table_Spec.js
```diff
@@ -0,0 +1,108 @@
...
+    var triangleGLTF = minimalTriangleGLTF;
+    var sharedBuffer = Buffer.from('abcdefghijklmnopqrstuvwxyz');
+    var batchTableAttributes = {
+        aToLInclusive: {
+            name: 'aToLInclusive',
```
lilleyse: I like this idea! Nice way to test it.

samples-generator/specs/bin/CESIUM_3dtiles_batch_table_Spec.js
```diff
@@ -0,0 +1,108 @@
...
+        expect(lastBufferView.byteLength).toBe(batchTableAttributes.mToZInclusive.byteLength);
+        expect(lastBufferView.byteOffset).toBe(batchTableAttributes.mToZInclusive.byteOffset);
+
+        // verify that references to the binary accessors are in `extensions: {...}`
+        var batchTableProperties = gltfWithExt.extensions.CESIUM_3dtiles_batch_table.batchTables[0].properties;
+        expect(batchTableAttributes.aToLInclusive.name in batchTableProperties).toBeTrue();
+        expect(batchTableAttributes.mToZInclusive.name in batchTableProperties).toBeTrue();
+        expect(batchTableProperties[batchTableAttributes.aToLInclusive.name].accessor).toBe(oldAccessorLength);
+        expect(batchTableProperties[batchTableAttributes.mToZInclusive.name].accessor).toBe(oldAccessorLength + 1);
+    });
```
lilleyse: Excellent tests!

samples-generator/specs/bin/minimalTriangleGLTF.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+module.exports = {
```
lilleyse: I'd kinda rather this just be a file on disk, especially if we start adding more glTF test cases.

lilleyse: It could go in a new folder `specs/data`

samples-generator/lib/createBuildingsTile.js
Samulus: fixed, see `injectTranslationNodeToGltfSpec.js` for details on how it works now. 

The batched examples always have a rotationMatrix as the root, so I create a translation node as a child of the rotationMatrix, then move the `mesh: 0` property from the root rotationMatrix to it's own node (as the child of the translationMatrix)

Samulus:
> 
> 
> Few things here and there but the the batched tiles look like they're nearly ready.

@lilleyse All comments addressed

samples-generator/lib/createBuildingsTile.js
```diff
@@ -91,24 +104,57 @@ function createBuildingsTile(options) {
...
+    };
+
+    var binary = defined(b3dmOptions.batchTableBinary) ? b3dmOptions.batchTableJsonAndBinary.binary : undefined;
+
+    // Don't add the batch table extension if there is no batchTableJson (e.g in the case of `createBatchedWithoutBatchTable`)
+    if (use3dTilesNext && defined(b3dmOptions.batchTableJson)) {
+        gltf = create3dtilesBatchTableExt(gltf, b3dmOptions.batchTableJson, binary);
+
+        // Make the RTC_CENTER translation the child of the rotationMatrix rootNode
+        gltf = injectTranslationNodeToGltf(gltf, 'RTC_CENTER', featureTableJson.RTC_CENTER, 0);
```
lilleyse: `injectTranslationNodeToGltf` shouldn't be called if `featureTableJson.RTC_CENTER` is undefined. `BatchedWGS84` is one case like that.

samples-generator/lib/createBuildingsTile.js
```diff
@@ -91,24 +104,57 @@ function createBuildingsTile(options) {
...
+
+        // Make the root mesh: 0 node a child of the RTC_TRANSLATION node.
+        var value = gltf.nodes[0].mesh;
+        delete gltf.nodes[0].mesh;
+        gltf.nodes[1].children = [2];
+        gltf.nodes.push({mesh: value});
```
lilleyse: It might be simpler for the mesh to move to the RTC node instead of adding a new node.

samples-generator/lib/injectTranslationNodeToGltf.js
```diff
@@ -0,0 +1,79 @@
...
+ *          an element in the nodes tree that is definitely a root
+ *          (no other node claims it as a child).
+ */
+
+function getRootNodesInGltf(nodes) {
+    var i = 0;
+    var potentialRoot = {};
+    for (; i < nodes.length; ++i) {
+        potentialRoot[i] = true;
+    }
```
lilleyse: Style:

```suggestion
    var i;
    var potentialRoot = {};
    for (i = 0; i < nodes.length; ++i) {
        potentialRoot[i] = true;
    }
```

samples-generator/lib/injectTranslationNodeToGltf.js
```diff
@@ -0,0 +1,79 @@
...
+                 .map(function(v) { return parseInt(v); })
+                 .sort();
+}
+
+/**
+ * Inserts a translation node at the root or underneath the specified
+ * parentIndex if provided.
+ * @param {Object} gltf The GLTF asset to modify
+ * @param {Array<number>} translationVector XYZ translation value
+ * @param {Number=} parentIndex Optional parent index for the new node.
```
lilleyse:
```suggestion
 * @param {Number} parentIndex Optional parent index for the new node.
```

samples-generator/lib/injectTranslationNodeToGltf.js
```diff
@@ -0,0 +1,79 @@
...
+ *                              The injected translation node will become the sole child
+ *                              of the provided parent, and orphaned children
+ *                              will be 'adopted' by the injected translation node.
+ *                              of being placed at the root.
```
lilleyse: Typo?

samples-generator/lib/injectTranslationNodeToGltf.js
```diff
@@ -0,0 +1,79 @@
...
+    return Object.keys(potentialRoot)
+                 .map(function(v) { return parseInt(v); })
+                 .sort();
+}
+
+/**
+ * Inserts a translation node at the root or underneath the specified
+ * parentIndex if provided.
+ * @param {Object} gltf The GLTF asset to modify
+ * @param {Array<number>} translationVector XYZ translation value
```
lilleyse:
```suggestion
 * @param {Array.<Number>} translationVector XYZ translation value
```

samples-generator/lib/injectTranslationNodeToGltf.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
```
lilleyse: I think this solution is a bit overarchitected since the RTC node could just be manually inserted into the node hierarchy in `createGltf`. A utility like this is probably great for gltf-pipeline but I don't think this repo necessarily benefits from having it, unless we expect there to be multiple different use cases for it in here.

samples-generator/lib/saveJson.js
```diff
@@ -1,25 +1,33 @@
...
-function saveTilesetJson(path, json, prettyJson) {
+function saveJson(path, json, prettyJson, gzip) {
     prettyJson = defaultValue(prettyJson, true);
     var options = {};
     if (prettyJson) {
         options.spaces = 2;
     }
+    gzip = defaultValue(gzip, false);
+    if (gzip) {
+        return saveBinary(path, json, gzip);
```
lilleyse: `json` will need to get converted to a buffer first.

samples-generator/lib/createBuildingsTile.js
lilleyse: Also see the other comment about moving more node hierarchy changes to `createGltf`. This code already has built-in assumptions about the node hierarchy, it would make more sense to make these changes when the node hierarchy is being formed.

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -926,24 +950,59 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
...
+    var tilesetPath = path.join(tilesetDirectory, 'tileset.json');
+
     return createBuildingsTile(tileOptions)
         .then(function(result) {
+            if (argv['3d-tiles-next']) {
+                tilesetOptions.versionNumber = 1.1;
+
+                if (argv.glb) {
+                    if (tilesetOptions.contentDataUri) {
+                        tilesetOptions.contentUri = 'data:model/gltf-binary;base64,' + Buffer.from(result).toString('base64');
```
lilleyse: The tile should not be saved as a separate file if it's being stored as a data uri in `tileset.json`.

Selection_468

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -913,8 +937,8 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
             tilesetOptions.properties = getProperties(batchTableJson);
             var tilesetJson = createTilesetJsonSingle(tilesetOptions);
             var promises = [
-                saveTilesetJson(tilesetPath, tilesetJson, prettyJson),
-                saveTile(tilePath, i3dm, gzip)
+                saveJson(tilesetPath, tilesetJson, prettyJson),
```
lilleyse: Every call to `saveJson` should be passing through the gzip option.

samples-generator/lib/create3dtilesBatchTableExt.js
```diff
@@ -0,0 +1,137 @@
...
+ * @param {Object} gltf The gltf object to edit
+ * @param {batchTableAttribute} batchTableAttributes List of batch table attributes. The function will intelligently try to detect
+ *                                                   if the attribute is a human readable value (a key value where the value is
+ *                                                   just an array of non-binary data) or if the values are more keys describing the
+ *                                                   layout of binary data in a binary buffer.
+ * @param {Buffer}              sharedBinaryBuffer   Binary buffer. Must be defined if using at least 1 binary batch attribute.
+ *                                                   This function currently assumes that all of the binary batch attributes in
+ *                                                   batchTableAttributes are directly referring to this buffer.
+ */
+function create3dtilesBatchTableExt(gltf, batchTableAttributes, sharedBinaryBuffer) {
```
lilleyse: Did you see this comment: https://github.com/CesiumGS/3d-tiles-validator/pull/186/files#r386654358?

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -926,24 +950,59 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
...
+                return Promise.all([
+                    saveJson(tilesetPath, createTilesetJsonSingle(tilesetOptions), prettyJson),
+                    saveJson(tilePath, result, prettyJson)
+                ]);
+            }
+
+            // old .b3dm
             var b3dm = result.b3dm;
             var batchTableJson = result.batchTableJson;
             tilesetOptions.properties = getProperties(batchTableJson);
```
lilleyse: Did you see this comment: https://github.com/CesiumGS/3d-tiles-validator/pull/186/files#r386655831?

samples-generator/lib/createBuildingsTile.js
```diff
@@ -74,10 +86,11 @@ function createBuildingsTile(options) {
             var batchTableExtra = generateBatchTableExtra(buildings);
             batchTableJson = combine(batchTableJson, batchTableExtra);
         }
+
         if (createBatchTableBinary) {
-            var batchTableJsonAndBinary = generateBatchTableBinary(buildings);
-            batchTableJson = combine(batchTableJson, batchTableJsonAndBinary.json);
+            batchTableJsonAndBinary = (use3dTilesNext) ? generateBatchTableBinary3dTilesNext(buildings) : generateBatchTableBinary(buildings);
```
lilleyse:
```suggestion
            batchTableJsonAndBinary = use3dTilesNext ? generateBatchTableBinary3dTilesNext(buildings) : generateBatchTableBinary(buildings);
```

samples-generator/lib/injectTranslationNodeToGltf.js
Samulus: Agreed. I removed `injectTranslationNodeToGltf.js` and it's associated unittest.

samples-generator/lib/injectTranslationNodeToGltf.js
Samulus: deleted file, typo irrelevant now

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -926,24 +950,59 @@ function saveInstancedTileset(tilesetName, tileOptions, tilesetOptions) {
...
     tileOptions.transform = defaultValue(tileOptions.transform, buildingsTransform);
     tileOptions.relativeToCenter = defaultValue(tileOptions.relativeToCenter, true);
-
     tilesetOptions = defaultValue(tilesetOptions, {});
+
+    var ext = '';
+    if (argv['3d-tiles-next']) {
+        tileOptions.use3dTilesNext = true;
+        tileOptions.useGlb = argv.glb;
+        tileOptions.versionNumber = 1.1;
```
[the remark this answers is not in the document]
Samulus: Yes it can be removed: Side note, `tilesetOptions.versionNumber = 1.1` should be `'1.1'` instead according to the tileset spec. I've fixed that.

samples-generator/bin/3d-tiles-samples-generator.js
```diff
@@ -2781,172 +2840,342 @@ function createRequestVolume() {
                     transform : buildingTransform,
                     boundingVolume : {
                         box : buildingBoxLocal
+
                     },
+
                     geometricError : 0.0,
+
```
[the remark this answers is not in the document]
Samulus: Yeah no idea what happened here, my formatter must have gone crazy. Fixed

samples-generator/bin/3d-tiles-samples-generator.js
Samulus: fixed by moving 
```
var batchTableJson = result.batchTableJson;
var batchTableJson = result.batchTableJson;
tilesetOptions.properties = getProperties(batchTableJson);
```
higher in the function, so that properties is initialized by the time the gltf / glb / tileset with embedded gltf / glb is written.

samples-generator/lib/createBuildingsTile.js
Samulus: fixed this and every other instance of a ternary condition missing parens, check with the regex: `=\s*([^(].*[^)]) \? (.*) : (.*)`
It's not a perfect regex, it breaks on === etc, but if you use it in the `search` tab in vscode you can preview everywhere that `(=|==|=== cond ? a : b)` occurs

samples-generator/bin/3d-tiles-samples-generator.js
Samulus: fixed, BatchedWithContentDataUri doesn't spit out a gltf / glb / b3dm anymore

samples-generator/lib/saveJson.js
```diff
@@ -1,25 +1,32 @@
...
-function saveTilesetJson(path, json, prettyJson) {
+function saveJson(path, json, prettyJson, gzip) {
     prettyJson = defaultValue(prettyJson, true);
     var options = {};
     if (prettyJson) {
         options.spaces = 2;
     }
+    gzip = defaultValue(gzip, false);
+    if (gzip) {
+        return saveBinary(path, Buffer.from(json), gzip);
```
lilleyse:
```suggestion
        return saveBinary(path, Buffer.from(JSON.stringify(json)), gzip);
```

If gzip is true it crashes without this fix.

samples-generator/lib/createBuildings.js
```diff
@@ -124,17 +124,18 @@ function createRandomBuildings(options) {
     // Set the random number seed before creating materials
     CesiumMath.setRandomNumberSeed(seed);
     var materials = new Array(numberOfBuildings);
+    var i;
     for (i = 0; i < numberOfBuildings; ++i) {
         // For CesiumJS testing purposes make the first building red
         var useRedMaterial = (baseColorType === 'color') && (translucencyType === 'opaque') && i === 0;
         var randomMaterial = getMaterial(baseColorType, translucencyType, i, numberOfBuildings);
-        materials[i] = useRedMaterial ? redMaterial : randomMaterial;
+        materials[i] = (useRedMaterial) ? redMaterial : randomMaterial;
```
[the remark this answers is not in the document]
Samulus: leaving it as-is then

Samulus:
@lilleyse Feedback addressed, there's still the issue of all gltfs passing validation but some glbs failing.

lilleyse:
@Samulus I made a few last minute changes:

* Brought in the gltf-pipeline fix: https://github.com/CesiumGS/gltf-pipeline/pull/528. All samples pass validation now, gltf and glb.
* Removed `nestingBoxes.gltf`
* Minor whitespace changes

lilleyse:
All looks good to me now!

PR #188, opened 2020-03-04
Don't bake scale into geometric error in discrete lod example

lilleyse:
For https://github.com/CesiumGS/cesium/issues/8660


Samulus:
👍 Can confirm this fixes the LOD issue, tested locally.

PR #194, opened 2020-03-19
Khr mesh instancing

Samulus:


lilleyse:
Just some general thoughts from inspecting the output:

* `InstancedTextured` and `InstancedWithTransform` no longer get generated. Is that intentional?
* `InstancedOrientation` has feature ids / feature table whereas other examples like `InstancedScale` don't. The legacy code puts feature tables in everything except `InstancedWithoutBatchTable`. However, now that we have freedom to change things I think we should only put feature tables in `InstancedWithBatchTable` and `InstancedWithBatchTableBinary`. `InstancedWithoutBatchTable` may be obsolete now because there would be plenty of other examples without a feature table. (This also reminds me, we need to update the names of the generated samples soon... this is starting to get confusing haha).
* `_FEATURE_ID_0` is in the `CESIUM_3dtiles_feature_metadata` object but there's no `_FEATURE_ID_0` in the primitive's attributes. In the older version of the feature metadata spec this would be a bug but let's forgo fixing that and just start transitioning the samples to the latest spec which would look like the first example [here](https://github.com/CesiumGS/glTF/tree/batch-table/extensions/2.0/Vendor/CESIUM_3dtiles_feature_metadata#implicit-feature-ids).
* `InstancedWithBatchTableBinary` - the legacy version only uses the `id` property and doesn't mix json and binary properties. It might make everything simpler though if only `Height` existed and `InstancedWithBatchTable` used `values` and `InstancedWithBatchTableBinary` used `accessor`.

lilleyse:
Did a quick pass of the code. Nothing major to comment on, overall very pleased with the new direction the code is taking.

Pushed a small typo fix: https://github.com/CesiumGS/3d-tiles-validator/pull/194/commits/d9ed12036d914d12734a575936ece420fcc2b1cd

Samulus:
@lilleyse Everything addressed, but you might want to take a look at the featureTables section for `InstancedWithBatchTable` / `InstancedWithBatchTableBinary` to be sure it matches the latest version of the spec.

`InstancedWithBatchTable` has 24 heights, each of them equal to 20. It's a candidate for the implicit feature of the latest version of the spec.

lilleyse:
> `InstancedWithBatchTable` has 24 heights, each of them equal to 20. It's a candidate for the implicit feature of the latest version of the spec.

The current approach is not quite right since it assigns each instance the feature id 20, but this is a feature id not the actual height value. The approach should be to assign feature id 0 to all instances and have a `height` property in the feature table with a single value.

All that said, I think we should keep it simple for now and let each instance have its own height value in case we decide to randomize the heights later.

```json
"extensions": {
  "CESIUM_3dtiles_feature_metadata": {
    "featureLayers": [
      {
        "featureTable": 0,
        "instanceStride": 1,
        "vertexAttribute": {
          "implicit": {
            "start": 0,
            "increment": 0
          }
        }
      }
    ]
  }
}
```

```json
  "extensions": {
    "CESIUM_3dtiles_feature_metadata": {
      "featureTables": [
        {
          "featureCount": 24,
          "properties": {
            "Height": {
              "values": [20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20]
            }
          }
        }
      ]
    }
  }
```

`InstancedWithBatchTableBinary` should look similar, except that `Height` should be stored in an accessor rather than values. From above:

> It might make everything simpler though if only `Height` existed and `InstancedWithBatchTable` used `values` and `InstancedWithBatchTableBinary` used `accessor`.

Samulus:
@lilleyse Fixed, also I added jest / istanbul support I can back out the commit responsible (`15a86cdc`) into a separate PR if you want.

lilleyse:
@Samulus don't forget `"instanceStride": 1,` in the example I posted above.

Also please update `InstancedWithBatchTableBinary` to match `InstancedWithBatchTable` as I mentioned above, with the only difference being `Height` stored as an accessor instead of values. I don't think the primitive should be using explicit feature ids. The feature table part looks good, just update the primitive.

**Current**
```
      "primitives": [
        {
          "attributes": {
            "POSITION": 1,
            "NORMAL": 2,
            "_FEATURE_ID_0": 4
          },
          "indices": 0,
          "material": 0,
          "mode": 4,
          "extensions": {
            "CESIUM_3dtiles_feature_metadata": {
              "featureLayers": [
                {
                  "featureTable": 0,
                  "vertexAttribute": {
                    "attributeindex": 0
                  }
                }
              ]
            }
          }
        }
      ]
```

**Expected**
```
      "primitives": [
        {
          "attributes": {
            "POSITION": 1,
            "NORMAL": 2
          },
          "indices": 0,
          "material": 0,
          "mode": 4,
          "extensions": {
            "CESIUM_3dtiles_feature_metadata": {
              "featureLayers": [
                {
                  "featureTable": 0,
                  "instanceStride": 1,
                  "vertexAttribute": {
                    "implicit": {
                      "increment": 0,
                      "start": 0
                    }
                  }
                }
              ]
            }
          }
        }
      ]
```

samples-generator/lib/attribute.ts
```diff
@@ -0,0 +1,13 @@
...
+    buffer: Buffer;
+    propertyName: string;
+    byteOffset?: number;
+    byteAlignment?: number;
+    componentType: GLenum;
+    count: number;
+    max: number[];
+    min: number[];
+    type: GltfType;
+}
```
sanjeetsuhag, line 13: Some files have different line endings. This may not be a concern with TS, but just wanna be sure that you don't miss this, in case we need to keep them consistent.

samples-generator/lib/instanceUtilsNext.ts
```diff
@@ -0,0 +1,373 @@
...
+const Cartesian3 = Cesium.Cartesian3;
+const CesiumMath = Cesium.Math;
+
+export namespace InstanceTileUtils {
+    export interface FeatureTableJson {
+        INSTANCES_LENGTH?: number;
+        RTC_CENTER?: [number, number, number];
+        QUANTIZED_VOLUME_SCALE?: [number, number, number];
+        QUANTIZED_VOLUME_OFFSET?: [number, number, number];
+        EAST_NORTH_UP?: boolean;
```
sanjeetsuhag, line 17: This is not being handled right now, but looks like it might need to be: [Refer to Sean's comment on my PR](https://github.com/CesiumGS/swayze/pull/1639#discussion_r403504035)

samples-generator/lib/attribute.ts
Samulus: Are you talking about the trailing whitespace in each file?  All the files show up as `LF` line ending for me. Prettier automatically inserts a trailing newline, I don't automatically run prettier on every file though (because it would format parts of existing code that then show up as irrelevant changes in a pull request).

samples-generator/lib/createEXTMeshInstancing.ts
```diff
@@ -0,0 +1,21 @@
+import { Gltf, GltfNode } from './gltfType';
+import { KHRMeshInstancing } from './khrInstancingType';
+const extensionName = 'EXT_mesh_gpu_instancing';
+
+export function createEXTMeshInstancingExtension(
+    gltf: Gltf,
+    node: GltfNode,
+    khrMeshInstancing: KHRMeshInstancing
+) {
+    if (gltf.extensionsUsed == null) {
```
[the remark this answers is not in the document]
Samulus: fixed

samples-generator/lib/featureMetadata.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Gltf, GltfPrimitive } from './gltfType';
+import { FeatureTable, FeatureLayer } from './featureMetadataType';
+
+export namespace FeatureMetadata {
+    /**
+     * Updates the extensionsUsed section of a provided glTF asset. Creates
+     * this array if it does not exist already.
+     * @param gltf The glTF asset to modify
+     */
+    export function updateExtensionUsed(gltf: Gltf) {
```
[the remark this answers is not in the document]
Samulus: fixed

samples-generator/lib/createEXTMeshInstancing.ts
```diff
@@ -0,0 +1,21 @@
+import { Gltf, GltfNode } from './gltfType';
+import { KHRMeshInstancing } from './khrInstancingType';
+const extensionName = 'EXT_mesh_gpu_instancing';
+
+export function createEXTMeshInstancingExtension(
+    gltf: Gltf,
+    node: GltfNode,
+    khrMeshInstancing: KHRMeshInstancing
```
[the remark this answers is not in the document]
Samulus: changed all references of khr_mesh_instancing to EXT…

samples-generator/lib/attribute.ts
Samulus: I ran the formatter on everything new, but if you see anything else bad style-wise let me know.

sanjeetsuhag: Your changes look good to me.

samples-generator/lib/instanceUtilsNext.ts
Samulus: Everything but `createInstancedWithTransform` and `createInstancedOrientation` now use EAST_NORTH_UP, see changes to `instanceUtilsNext.ts`

sanjeetsuhag:
@Samulus All the samples look good! Thanks!

PR #195, opened 2020-03-27
Composite samples

Samulus:
WIP, needs to be rebased after KHR_mesh_instancing is updated, but the actual outputted samples are reviewable now. (See `CompositeOfInstanced` and `Composite`)


samples-generator/lib/createFeatureMetadataExtension.ts
```diff
@@ -0,0 +1,271 @@
...
+
+            const implicitBufferView = isCandidateForImplicitBufferView(
+                batchAttribute
+            );
+            if (!implicitBufferView) {
+                gltf.bufferViews.push({
+                    buffer: newBufferIndex,
+                    byteLength: batchAttribute.byteLength,
+                    byteOffset: batchAttribute.byteOffset,
+                    target: 0x8892 // ARRAY_BUFFER
```
sanjeetsuhag, line 234: Maybe in the next pass, you could create an explicit Type for this too. Just to make it more readable.

Samulus:
@sanjeetsuhag all comments addressed

lilleyse:
Thanks @Samulus!

PR #209, opened 2020-10-23
Validate 3DTILES_content_gltf extension

lilleyse:
This PR adds support for the `3DTILES_content_gltf` extension which allows a tileset to refer to glTF and glb content directly.
There's still backwards compatible support for tilesets with version `'2.0.0-alpha.0'`

See https://github.com/CesiumGS/3d-tiles/pull/427 for the in-progress spec.

### DIFF
--- a/validator/lib/isTile.js
+++ b/validator/lib/isTile.js
@@ -5,7 +5,7 @@ const defaultValue = Cesium.defaultValue;
 const defined = Cesium.defined;
 module.exports = isTile;
 
-var emptyArray = [];
+const EMPTY_ARRAY = [];
 
 /**
  * Checks whether the given file path is a tile file path.
@@ -15,19 +15,19 @@ var emptyArray = [];
  * @returns {Boolean} True if the file path is a tile file path, false if not.
  */
 function isTile(filePath, options) {
-    var extension = path.extname(filePath);
+    const extension = path.extname(filePath);
     if (defined(options)) {
-        var version = options.version;
-        var tileset = options.tileset;
-        var extensionsUsed = defaultValue(tileset.extensionsUsed, emptyArray);
-        var hasContentGltfExtension = extensionsUsed.indexOf('3DTILES_content_gltf') > -1;
+        const version = options.version;
+        const tileset = options.tileset;
+        const extensionsUsed = defaultValue(tileset.extensionsUsed, EMPTY_ARRAY);
+        const hasContentGltfExtension = extensionsUsed.indexOf('3DTILES_content_gltf') > -1;
 
         // Version "2.0.0-alpha.0" was used during early development of 3D Tiles Next
         // and is included here for backwards compatibility purposes
         if (version === '2.0.0-alpha.0' || hasContentGltfExtension) {
             return (
                 extension === '.gltf' ||
-                extension === '.glb' || 
+                extension === '.glb' ||
                 extension === '.b3dm' ||
                 extension === '.i3dm' ||
                 extension === '.pnts' ||
@@ -38,15 +38,14 @@ function isTile(filePath, options) {
                 extension === '.i3dm' ||
                 extension === '.pnts' ||
                 extension === '.cmpt');
-        } else {
-            return false;
         }
+        return false;
     }
 
     // if no version specified, match any
     return (
         extension === '.gltf' ||
-        extension === '.glb' || 
+        extension === '.glb' ||
         extension === '.b3dm' ||
         extension === '.i3dm' ||
         extension === '.pnts' ||
--- a/validator/lib/isTile.js
+++ b/validator/lib/isTile.js
@@ -22,6 +22,8 @@ function isTile(filePath, options) {
         var extensionsUsed = defaultValue(tileset.extensionsUsed, emptyArray);
         var hasContentGltfExtension = extensionsUsed.indexOf('3DTILES_content_gltf') > -1;
 
+        // Version "2.0.0-alpha.0" was used during early development of 3D Tiles Next
+        // and is included here for backwards compatibility purposes
         if (version === '2.0.0-alpha.0' || hasContentGltfExtension) {
             return (
                 extension === '.gltf' ||
--- a/validator/lib/isTile.js
+++ b/validator/lib/isTile.js
@@ -1,29 +1,41 @@
 'use strict';
 const path = require('path');
 const Cesium = require('cesium');
+const defaultValue = Cesium.defaultValue;
 const defined = Cesium.defined;
 module.exports = isTile;
+
+var emptyArray = [];
 
 /**
  * Checks whether the given file path is a tile file path.
  *
  * @param {String} filePath The file path.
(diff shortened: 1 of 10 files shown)


validator/lib/isTile.js
```diff
@@ -1,29 +1,41 @@
...
     var extension = path.extname(filePath);
-    if (defined(version)) {
-        if (version === '1.0') {
+    if (defined(options)) {
+        var version = options.version;
+        var tileset = options.tileset;
+        var extensionsUsed = defaultValue(tileset.extensionsUsed, emptyArray);
+        var hasContentGltfExtension = extensionsUsed.indexOf('3DTILES_content_gltf') > -1;
+
+        if (version === '2.0.0-alpha.0' || hasContentGltfExtension) {
```
sanjeetsuhag, line 25: Is the `2.0.0-alpha.0` version documented anywhere? It's not in the 3DTILES_content_gltf PR.

lilleyse: I added a clarifying comment: https://github.com/CesiumGS/3d-tiles-validator/pull/209/commits/975fab4ac80cd25ec42d9cc6817a5c4488689b1e

validator/lib/isTile.js
```diff
@@ -1,29 +1,43 @@
...
-function isTile(filePath, version) {
+function isTile(filePath, options) {
     var extension = path.extname(filePath);
-    if (defined(version)) {
-        if (version === '1.0') {
+    if (defined(options)) {
+        var version = options.version;
+        var tileset = options.tileset;
+        var extensionsUsed = defaultValue(tileset.extensionsUsed, emptyArray);
+        var hasContentGltfExtension = extensionsUsed.indexOf('3DTILES_content_gltf') > -1;
```
sanjeetsuhag, line 23: This marks a tileset as valid if `3DTILES_content_gltf` is in `extensionsUsed` and not `extensionsRequired`. I'm assuming any tileset that needs this extension will need to add it to the `extensionsRequired`, correct?

lilleyse: Hm this makes me wonder...

[3DTILES_content_gltf](https://github.com/CesiumGS/3d-tiles/tree/3DTILES_content_gltf/extensions/3DTILES_content_gltf) says it needs to go in `extensionsRequired` but now I'm thinking about the interaction with [3DTILES_layers](https://github.com/CesiumGS/3d-tiles/pull/432) and how someone might create a tileset with both b3dm and glTF layers so that the b3dm layer can be a fallback if the client doesn't support `3DTILES_content_gltf`. That's a possible use case for allowing `3DTILES_content_gltf` to be optional.

lilleyse: I left a comment in https://github.com/CesiumGS/3d-tiles/pull/427#issuecomment-715598362.

lilleyse: For this area of the code it's ok to just check `extensionsUsed` since the extension will always go in `extensionsUsed` whether it's required or not.

I pushed some changes that do more thorough checking of the behavior of `3DTILES_content_gltf`. Also went through and fixed some broken unit tests and eslint errors.

validator/lib/validateExtensions.js
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+    }
+}
+
+function validate3DTilesContentGltf(extension) {
+    const gltfExtensionsUsed = defaultValue(extension.extensionsUsed, EMPTY_ARRAY);
+    const gltfExtensionsRequired = defaultValue(extension.extensionsRequired, EMPTY_ARRAY);
+
+    if (!Array.isArray(gltfExtensionsUsed)) {
```
sanjeetsuhag, line 53: I think we can simplify this further using:

```javascript
if (!Array.isArray(gltfExtensionsUsed) || gltfExtensionsUsed.filter(i => typeof i === 'string').length) {
  return 'extensionsUsed must be an array of strings';
}
```

validator/lib/validateExtensions.js
```diff
@@ -0,0 +1,72 @@
...
+        return 'extensionsUsed must be an array of strings';
+    }
+
+    if (!Array.isArray(gltfExtensionsRequired)) {
```
sanjeetsuhag, line 57: Same as [above](https://github.com/CesiumGS/3d-tiles-validator/pull/209/commits/6f1993e86615c77589565f159a31c36e97ac6aee#r511199787):
```javascript
if (!Array.isArray(gltfExtensionsRequired) || !gltfExtensionsRequired.filter(i => typeof i === 'string').length) {
  return 'extensionsRequired must be an array of strings';
}
```

sanjeetsuhag (commented):
~~I just have two small comments above making the code less verbose for the simple stuff, but otherwise looks good.~~

Upon further review, changes are needed.

validator/lib/validateExtensions.js
```diff
@@ -0,0 +1,72 @@
...
+    const tileset = options.tileset;
+    const extensions = tileset.extensions;
+    const extensionsUsed = defaultValue(tileset.extensionsUsed, EMPTY_ARRAY);
+    const extensionsRequired = defaultValue(tileset.extensionsRequired, EMPTY_ARRAY);
+
+    if (defined(extensions)) {
+        for (const extensionName in extensions) {
+            if (extensions.hasOwnProperty(extensionName)) {
+                const extension = extensions[extensionName];
+                if (!extensionsUsed.includes(extensionName)) {
```
sanjeetsuhag, line 30: We need to check if `extensionsUsed` and `extensionsRequired` are array before we call `.includes()`.

validator/lib/validateExtensions.js
```diff
@@ -0,0 +1,72 @@
...
+                }
+                if (defined(message)) {
+                    return `Error in ${extensionName}: ${message}`;
+                }
+            }
+        }
+    }
+}
+
+function validate3DTilesContentGltf(extension) {
```
sanjeetsuhag, line 49: Three's nothing specific to `3DTILES_content_gltf` here, this code should be moved above into the loop before we call `.includes()`.

lilleyse: This code is checking the inner glTF `extensionUsed` and `extensionsRequired`

```json
"extensionsUsed": ["3DTILES_content_gltf"],
"extensionsRequired": ["3DTILES_content_gltf"],
"extensions": {
  "3DTILES_content_gltf": {
    "extensionsUsed": ["KHR_texture_transform"],
    "extensionsRequired": ["KHR_texture_transform"],
  }
}
```

validator/lib/validateExtensions.js
lilleyse: Fixed

validator/lib/validateExtensions.js
lilleyse: I refactored the code a bit so everything calls `validateExtensionsArray`. I kept the code similar to before for readability but now it only happens in one place.

lilleyse:
Replaced by #222

PR #212, opened 2020-12-16
b3dm file sizes should be a multiple of 8 bytes.

renjianqiang:
Fixes CesiumGS/3d-tiles-samples-generator#1
modify the samples-generator and validator to let b3dm file sizes(`byteLength`)  aligned to an 8-byte boundary.

### DIFF
--- a/samples-generator/lib/createB3dm.js
+++ b/samples-generator/lib/createB3dm.js
@@ -21,9 +21,9 @@ module.exports = createB3dm;
  * @returns {Buffer} The generated b3dm tile buffer.
  */
 function createB3dm(options) {
-    var glb = options.glb;
+    var glb = getBufferPadded(options.glb);
     var defaultFeatureTable = {
-        BATCH_LENGTH : 0
+        BATCH_LENGTH: 0
     };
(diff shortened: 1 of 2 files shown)


renjianqiang:
I also find same problem in `i3dm` , i will fix it if allowed.

lilleyse:
@javagl does the new validator handle these checks properly? If so this PR can be closed.

The samples-generator fix would need to be applied to https://github.com/CesiumGS/3d-tiles-samples-generator

javagl:
The B3DM validation was created by porting the state at https://github.com/CesiumGS/3d-tiles-validator/blob/e84202480eb6572383008076150c8e52c99af3c3/validator/lib/validateB3dm.js , which does not include this check. It can be added based on this PR, though.

javagl:
The changes in this PR that affected the validator are now part of https://github.com/CesiumGS/3d-tiles-validator/pull/223, so I'll close it here.

The changes that affect the samples generator could still be extracted from here, to fix https://github.com/CesiumGS/3d-tiles-samples-generator/issues/1

PR #217, opened 2021-12-13
Fixed param problem with `combine` script

### LINKED ISSUE #214: running combine
“The ‘path’ argument must be of type string. Received undefined”
i dont konw why does this problem occur。
image

getinthedamnbox:
Fixes #214 

Fixes problem with `combine` script: https://github.com/CesiumGS/3d-tiles-validator/issues/214

The problem is that the script requires `-r` but then tries to use `-i` to find the root JSON tile.

A user requested this fix: https://community.cesium.com/t/3d-tiles-specifying-tileset-json-pointing-to-external-tilesets/15225/5


ptrgags:
Change looks good to me, thanks @getinthedamnbox!

PR #219, opened 2022-05-06
Add ThirdParty.json

lilleyse:
Adds `ThirdParty.json` in each of the project folders:
* `validator/ThirdParty.json`
* `tools/ThirdParty.json`
* `samples-generator/ThirdParty.json`

To regenerate the file run `npm run generate-third-party` in each project folder

CC @shehzan10

### DIFF
--- a/samples-generator/ThirdParty.json
+++ b/samples-generator/ThirdParty.json
@@ -1,0 +1,58 @@
+[
+  {
+    "name": "bluebird",
+    "license": [
+      "MIT"
+    ],
+    "version": "3.7.2",
+    "url": "https://www.npmjs.com/package/bluebird"
+  },
+  {
+    "name": "cesium",
+    "license": [
+      "Apache-2.0"
+    ],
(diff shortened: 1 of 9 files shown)


samples-generator/ThirdParty.json
```diff
@@ -0,0 +1,58 @@
+[
+  {
+    "name": "bluebird",
```
shehzan10, line 3: @lilleyse We shouldn't list devdependencies as part of the thirdparty JSON since we don't use it in the shipped version. According to the package.json in this folder, Bluebird (and other libraries) are dev dependencies, right? Can you please check it in the other folders too?

lilleyse: `bluebird` is listed under `dependencies`: https://github.com/CesiumGS/3d-tiles-validator/blob/e4415ccb52595e29c43d575ef29507b343ad85e8/samples-generator/package.json#L25-L36

PR #221, opened 2022-07-06
Upgrades to eslint-cesium-config/node v9.0.0

sanjeetsuhag:


ggetz:
@sanjeetsuhag [The `README`](https://github.com/CesiumGS/3d-tiles-validator/tree/main/tools) says that the `Tools` directory is not being actively maintained. Do the other directories (`validator` and `samples-generator`) need these updates instead?

Issue #204, opened 2020-07-16
Report when region boundingVolumes are < -Pi or > Pi


jerstlouis:
I would naturally expect latitude & longitude to be specified in degrees as it is the most commonly used human-friendly unit.
The behavior of wrong bounding volumes is very puzzling (e.g. things disappear when zooming in, no console output in CesiumJS), so it makes it very difficult to debug what could be a common mistake.

It would go a long way to warn about region bounding volumes mistakenly specified as degrees by checking against the -Pi/2..Pi and -Pi..Pi ranges.

javagl:
This should have been closed (see https://github.com/CesiumGS/3d-tiles-validator/pull/222#issuecomment-1265905466 ).

Having an invalid bounding region like this
```
{
  "asset": {
    "version": "1.1"
  },
  "geometricError": 4.0,
  "root": {
    "boundingVolume": {
      "region": [
        181,
        91,
        -181,
        -91,
        10000,
        9999
      ]
    },
    "geometricError": 2.0
  }
}
```
will now generate issues like these
```
Validation result:
{
  "date": "2022-10-04T14:58:28.480Z",
  "numErrors": 6,
  "numWarnings": 0,
  "issues": [
    {
      "type": "BOUNDING_VOLUME_INCONSISTENT",
      "path": "/root/boundingVolume/region/0",
      "message": "The 'west' entry of the bounding region must be in [-180,180], but is 181",
      "severity": "ERROR"
    },
    {
      "type": "BOUNDING_VOLUME_INCONSISTENT",
      "path": "/root/boundingVolume/region/1",
      "message": "The 'south' entry of the bounding region must be in [-90,90], but is 91",
      "severity": "ERROR"
    },
    {
      "type": "BOUNDING_VOLUME_INCONSISTENT",
      "path": "/root/boundingVolume/region/2",
      "message": "The 'east' entry of the bounding region must be in [-180,180], but is -181",
      "severity": "ERROR"
    },
    {
      "type": "BOUNDING_VOLUME_INCONSISTENT",
      "path": "/root/boundingVolume/region/3",
      "message": "The 'north' entry of the bounding region must be in [-90,90], but is -91",
      "severity": "ERROR"
    },
    {
      "type": "BOUNDING_VOLUME_INCONSISTENT",
      "path": "/root/boundingVolume/region",
      "message": "The 'south' entry of the bounding region may not be larger than the 'north' entry, but the south is 91 and the north is -91",
[...]

Issue #126, opened 2017-08-25
Validator - check for required properties in tileset.json


lilleyse:
@rms13

`validateTileset` is doing a much better job now, but there are still a lot of loose ends. In the code verify that all required properties are present and are the correct type. This will be a sweeping PR to bring `validateTileset` to completion and will require looking the spec and tileset.json schema closely.

javagl:
This should have been closed via https://github.com/CesiumGS/3d-tiles-validator/pull/222#issuecomment-1265905466 

Having missing properties (like the missing `geometricError`) in a tileset like this
```
{
  "asset" : {
    "version" : "1.1"
  },
  "root" : {
    "boundingVolume" : {
      "box" : [ 0.5, 0.5, 0.5, 0.5, 0.0, 0.0, 0.0, 0.5, 0.0, 0.0, 0.0, 0.5 ]
    },
    "geometricError" : 1.0
  }
}
```
will now generate an issue like this
```
Validation result:
{
  "date": "2022-10-04T15:00:06.894Z",
  "numErrors": 1,
  "numWarnings": 0,
  "issues": [
    {
      "type": "PROPERTY_MISSING",
      "path": "/geometricError",
      "message": "The 'geometricError' property is required",
      "severity": "ERROR"
    }
  ]
}
```

The same should be the case for _all_ required filelds.

PR #222, opened 2022-09-21
Initial draft of 3D Tiles 1.1 validator

javagl:
This is an implementation of a validator for 3D Tiles 1.1. This new state has been developed in an internal repository, and is now moved here as a PR for first reviews. This state is supposed to replace the original validator. But it is **not** supposed to be merged before the `tools` and `samples-generator` projects from _this_ repository have been moved into new repositories.

The [README.md](https://github.com/javagl/3d-tiles-validator/blob/validator-1.1/README.md) shows the basic usage, based on the current state. There are some [discussion points in `IMPLEMENTATION.md`](https://github.com/javagl/3d-tiles-validator/blob/validator-1.1/IMPLEMENTATION.md#discussion-points) that will be addressed soon.


package.json
```diff
@@ -0,0 +1,36 @@
...
+    "eslint": "^8.0.1",
+    "eslint-config-cesium": "^9.0.0",
+    "eslint-config-prettier": "^8.3.0",
+    "jasmine": "^4.3.0",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4"
+  },
+  "scripts": {
+    "eslint": "eslint \"./**/*.*s\" --cache --quiet",
+    "build": "npx tsc",
```
lilleyse, line 32: I'm seeing some errors when running `npm run build`

```
(base) slilley@Hydrogen:~/Code/3d-tiles-validator$ npm run build

> 3d-tiles-validator-1.1@0.0.0 build
> npx tsc

src/validation/ValidationState.ts:1:24 - error TS2307: Cannot find module './structure/Metadata/Schema' or its 
corresponding type declarations.

1 import { Schema } from "./structure/Metadata/Schema";
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/validation/ValidationState.ts:2:23 - error TS2307: Cannot find module './structure/Group' or its corresponding type declarations.

2 import { Group } from "./structure/Group";
                        ~~~~~~~~~~~~~~~~~~~
```

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@types/jasmine": "^4.0.3",
+    "eslint": "^8.0.1",
+    "eslint-config-cesium": "^9.0.0",
+    "eslint-config-prettier": "^8.3.0",
+    "jasmine": "^4.3.0",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4"
+  },
+  "scripts": {
+    "eslint": "eslint \"./**/*.*s\" --cache --quiet",
```
lilleyse, line 31: `npm run eslint` isn't working out of the box for me

```
(base) slilley@Hydrogen:~/Code/3d-tiles-validator$ npm run eslint

> 3d-tiles-validator-1.1@0.0.0 eslint
> eslint "./**/*.*s" --cache --quiet


Oops! Something went wrong! :(

ESLint: 8.23.1

ESLint couldn't find the plugin "eslint-plugin-node".

(The package "eslint-plugin-node" was not found when loaded as a Node module from the directory "/home/slilley/Code/3d-tiles-validator".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-node@latest --save-dev

The plugin "eslint-plugin-node" was referenced from the config file in ".eslintrc.json » eslint-config-cesium/node".

If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.
```

IMPLEMENTATION.md
```diff
@@ -0,0 +1,98 @@
...
+  - `issues/SemanticValidationIssues.ts`: Issues related to inconsistent property values, like `TILE_GEOMETRIC_ERROR_INCONSISTENT` 
+  - `ContentValidationIssues.ts`: An error or warning in a tile content
+- For validation issues that refer to the tile content, each `ValidationIssue` can have an array of `internalIssues`. This can be filled, for example, with the information from the glTF validator that caused the validation to fail
+
+## Discussion points
+
+### General functionality
+
+- It should be possible to "batch process" a directory (with caveats - how to detect whether a file is a `tileset.json`, beyond its name?)
+- It should be possible to write reports into files.
```
lilleyse, line 60: +1 for both of these.

> how to detect whether a file is a `tileset.json`, beyond its name?

I think just checking the name is ok.

src/main.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Validators } from "./validation/Validators";
+const tilesSchemaRootDir = "C:/Develop/CesiumGS/3d-tiles/specification/schema";
+console.warn("Using fixed root directory for schema: " + tilesSchemaRootDir);
+Validators.setSchemaRootDir(tilesSchemaRootDir);
+//*/
+
+const args = yargs(process.argv.slice(1))
+  .help("help")
+  .alias("help", "h")
+  .options({
```
lilleyse, line 17: Feature request: ability to ignore warnings (for example: a tileset has a million glTF files that use an unrecognized extension)

src/structure/BoundingVolume.ts
```diff
@@ -0,0 +1,5 @@
+export interface BoundingVolume {
+  region?: number[];
+  box?: number[];
+  sphere?: number[];
```
lilleyse, line 4: Feature request: S2 support.

Otherwise certain tilesets don't validate (as expected)

```
    {
      "type": "ANY_OF_ERROR",
      "path": "/root/children/1/(level 0, (0,0))/(level 1, (1,0))/(level 2, (2,1))/(level 3, (5,2))/(level 4, (11,4))/boundingVolume",
      "message": "The 'boundingVolume' must define 'box', 'region', or 'sphere', but not does not define any of them",
      "severity": "ERROR"
    },
```

IMPLEMENTATION.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Can contain configuration options that may be needed in the future (like `doValidateContents=false` or so)
+- Maintains a `ResourceResolver` that reads a `Buffer` from a given URI (resolved in the respective context - the tileset, external tileset, or glTF)
+- Collects the `ValidationIssue` instances (errors and warnings) that are found during the traversal
+
+The `ValidationIssue` class and its types:
+
+- The `ValidationIssue` itself is a plain data structure with basic info about the issue (message, path, severity...)
+- Instances of this class are created with methods in the `issues*` classes, roughly categorized here:
+  - `issues/IoValidationIssues.ts`: Fundamental IO errors, like `JSON_PARSE_ERROR`
```
lilleyse, line 48: I got this error for one tileset. The `path` field is empty, is that expected? It almost looks like an internal error in the code that manifested as a validation error.

```
(base) slilley@Hydrogen:~/Code/3d-tiles-validator$ npx ts-node src/main.ts --tilesetFile ~/Desktop/CDB_yemen_gsmodel/tileset.json 
Validating tileset /home/slilley/Desktop/CDB_yemen_gsmodel/tileset.json
Validating as external tileset: GSModels_1_1.json
Validating as external tileset: Tiles/N12/E045/GSModels/1_1/N12E045_D300_S001_T001.json
Validating GLB: N12E045_D300_S001_T001_LC5_U0_R0.glb
Validating GLB: N12E045_D300_S001_T001_LC4_U0_R0.glb
Validating GLB: N12E045_D300_S001_T001_LC3_U0_R0.glb
Validating GLB: N12E045_D300_S001_T001_LC2_U0_R0.glb
Validating GLB: N12E045_D300_S001_T001_LC1_U0_R0.glb
Validation result:
{
  "date": "2022-09-21T20:14:46.871Z",
  "numErrors": 1,
  "numWarnings": 0,
  "issues": [
    {
      "type": "JSON_PARSE_ERROR",
      "path": "",
      "message": "TypeError: Cannot read properties of null (reading 'subarray')",
      "severity": "ERROR"
    }
  ]
}
```

lilleyse (commented):
Left some really early feedback that's usage related. Haven't dived into the code yet.

.gitignore
```diff
@@ -1,22 +1,3 @@
-# NPM
-node_modules/
-npm-debug.log
+node_modules
```
lilleyse, line 1: `build` should be added here as well (in case someone runs `npm run build`)

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "cesium": "^1.97.0",
+    "gltf-validator": "^2.0.0-dev.3.9",
+    "yargs": "^17.5.1"
+  },
+  "devDependencies": {
+    "@tsconfig/node16": "^1.0.3",
+    "@types/jasmine": "^4.0.3",
+    "eslint": "^8.0.1",
+    "eslint-config-cesium": "^9.0.0",
+    "eslint-config-prettier": "^8.3.0",
```
lilleyse, line 25: Is there a specific version of prettier you used to format the code? I think it should be added here and pinned to that version.

package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "3d-tiles-validator-1.1",
```
lilleyse, line 2: I would rename the package to `3d-tiles-validator`. Better name once we go to publish this on npm.

package-lock.json
```diff
@@ -1,3 +1,1499 @@
 {
```
lilleyse, line 1: Generally we don't check in `package-lock.json` for libraries. It can be removed.

src/tileFormats/legacy/defaultValue.js
```diff
@@ -0,0 +1,21 @@
+// From https://github.com/CesiumGS/wetzel
```
lilleyse, line 1: Duplicate with other `defaultValue.js`?

src/tileFormats/legacy/defined.js
```diff
@@ -0,0 +1,14 @@
+// From https://github.com/CesiumGS/wetzel
```
lilleyse, line 1: Duplicate with other `defined.js`?

src/tileFormats/legacy/utility.js
```diff
@@ -0,0 +1,57 @@
...
+        case 'FLOAT':
+            return 4;
+        case 'DOUBLE':
+            return 8;
+        default:
+            return undefined;
+    }
+}
+
+function isBufferValidUtf8(buffer){
```
lilleyse, line 54: This doesn't seem to be used anywhere. Same with `normalizePath`.

lilleyse:
Ok I dived into the code... but just barely. It looks good from what I've seen.

Some other validation checks that I didn't see in the code right away (correct me if wrong) include:

* Check that geometry is within `tile.boundingVolume`, `content.boundingVolume`, and `parent.boundingVolume`
* Check that metadata statistics are correct. Or at the very least, check that metadata values fall within min/max
* Check that metadata values fall within the class's min/max
* Declarative styling validation

These are somewhat more involved and don't necessarily need to hold up this PR. I'm sure others will come up over time.

package.json
javagl: The lack of refactoring functionalities of \*Script+VSCode is astonishing. Fixed this manually...

src/main.ts
javagl: This is covered in the IMPLEMENTATION.md -> Discussion Points -> Validation Options

src/structure/BoundingVolume.ts
javagl: This is largely covered by the IMPLEMENTATION.md -> Discussion Points -> JSON Schema Based Validation. 

The point there is to be able to add ~"something like" the S2 validation, based on a plain schema. 
The fact that this will affect the behavior of the existing validation is the main thing that makes this non-trivial (but I have some ideas about how that _might_ be solved). 

(This is, in fact, a case that is ~"somewhat structurally similar" to the question of how extensions may, for example, define additional valid `enum` values. Not entirely trivial, but certainly solvable)

IMPLEMENTATION.md
javagl: Using a plain name check _might_ be too narrow. Nobody knows what people might call their tilesets. The extreme case would be to allow the user (via config options) to define a regex for the filenames to be processed, with `tileset.json` being the default, or things like `tileset[.*].json` for things like `tileset_v1.json`.

But I consider this point to be on the radar due to the note in the IMPLEMENTATION.md

src/implicitTilingDemos.ts
```diff
@@ -0,0 +1,125 @@
+import path from "path";
```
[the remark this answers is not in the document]
javagl: Addressed in comment below

src/tileFormats/legacy/defaultValue.js
javagl: Addressed in comment below

package.json
javagl: I have fixed the linting to work for now (even though that involves ignoring the linting for some legacy code parts).

But there are several open questions about the configuration that are addressed below, and that I added as "High Level" discussion points in the implementation notes.

package.json
javagl: See comment below, and the updated implementation notes

IMPLEMENTATION.md
javagl: > I got this error for one tileset. The path field is empty, is that expected? It almost looks like an internal error in the code that manifested as a validation error.

Yes, it is an internal error that is not caught properly. I saw this once, with an implicit tileset where a part of the subtree handling was wrong (roughly: It tried to load a `subtree` file that was already detected to be invalid). 

Two notes:

- The implicit tiling part is not yet finalized, but the more test cases the better. If this is a tileset that is publicly/internally available, it would be good to run an own test. 
- Of course, these errors should appear as a `JSON_PARSE_ERROR`. (That's a too generic `catch(error)` at some point). I'll review these parts, and try to make sure that they only catch actual parsing errors (and report something like an `INTERNAL_ERROR` as a last resort - i.e. as something that should never happen, akin to an "assertion failure")

lilleyse:
You should be able to use the latest prettier version. CesiumJS uses `2.1.2` only because that was the latest version at the time prettier was added. A different, internal repo uses `2.6.2`. The important thing is that it's pinned to a specific version so that people don't get different formatting depending on which version of prettier is installed.

lilleyse:
> (Or I could just "port" the code to TypeScript. That might, in fact, be easier than working around the inconsistencies and glitches of such a mixed environment...)

Yeah, might be worth doing. It's just a few small files.

lilleyse:
Once the repo situation is sorted out it would be good to open separate issues for the discussion points in `IMPLEMENTATION.md`

IMPLEMENTATION.md
javagl: The fix for the _reason_ of the error was added in https://github.com/CesiumGS/3d-tiles-validator/pull/222/commits/caa14ab6856d8e548cbb0283947c2383a0d12f38 , together with some _highly_ preliminary error handling for similar cases in the future. The broader question about the handling of missing/unresolvable resources has been added to the implementation notes.

javagl:
Just a short update to summarize the latest changes:

- Added `prettier`, to be run with `num run prettier`
- Added _basic_ linting, to be run with `npm run eslint`. There are some details regarding to configuration. For example, `no-non-null-assertion` and `no-explicit-any` are currently disabled. Some settings _might_ make sense in the context of a validator that _has_ to anticipate that the input may be "broken" in an arbitrary way. Some others might just be caused by a lack of  understanding on my side, and where it could make sense to try to find "better" solutions.
- Updated the `tileFormats/legacy` to no longer be/use `.js` files
- Added a basic command line functionality for batch-processing tileset files from a directory. This refers to `*tileset*.json` files by default, but the matching pattern is an optional command line argument. 
- Cleanups, documentation, and refactoring for the implicit tileset traversal. The error handling is a bit difficult, but I hope that it is approaching a sensible solution now.
- Added basic Jasmine specs. They just validate all tileset- and metadata schema files from the `specs/data` directory, and check whether they generate the expected errors. The exact approach may be revised at some point. But for now, it at least serves the purpose of avoiding regressions (e.g. to notice that a certain issue is no longer detected after a refactoring step)

javagl:
A short update:

- Extension declarations are now validated. This means that it checks the `extensionsUsed` and `extensionsRequired` declarations, and all extensions that are found in the JSON, and generates appropriate issues. (E.g. an error when something is found in the JSON but not listed in `extensionsUsed`, and a warning when something is declared in `extensionsUsed` but not found in the JSON)
- With some hesitation, I made the validation functions return a `boolean`/`Promise<boolean>`. It's not sooo pretty to have to maintain the "validity state" during each function, and some changes (like turning an `ERROR` into a `WARNING` or vice versa) may require careful updates here. But from the caller perspective, it certainly makes sense.
- Zipped inputs are now unzipped and validated (on a per-file basis)
- A basic "report file" functionality has been added. The command line argument `writeReports` causes a `.report.json` file to be written for each validated file. For a single input, `reportFile` can be used to define the desired report file name
- The traversal+validation of implicit tilesets has been improved. In many cases that just reported an `IMPLICIT_TILING_ERROR` until now, it now validates the `.subtree` file _for each subtree root_, resulting in much more detailed issues
- Some bug fixes for GLB/I3DM handling: When the I3DM refers to a GLB via URI, the GLB is now also validated. The problems with the latter had been found by running against `cesium\Specs\Data\Cesium3DTiles`. There are still some issues that may be "false positives", but I'll regularly look into that.

lilleyse:
In `LICENSE.md` this sentence can be removed:

> All the tools are licensed with the Apache 2.0 license. Check the LICENSE.md in each directory for more information on third-party libraries.

Also can you update the copyright date?

javagl:
Merging this PR will fix some issues, or make them obsolete:

- Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/204 via [this check](https://github.com/javagl/3d-tiles-validator/blob/7e75f0cc828259efa36fb223e0e43868a9022b6b/src/validation/BoundingVolumeValidator.ts#L237)
- Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/126 via checks like [this one](https://github.com/javagl/3d-tiles-validator/blob/7e75f0cc828259efa36fb223e0e43868a9022b6b/src/validation/TilesetValidator.ts#L104)
- Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/107 via the `path` that is stored in each `ValidationIssue`
- Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/98 by calling the glTF validator on glTF content data

lilleyse:
I did a few final validation checks. Looks good!

Issue #227, opened 2022-10-04
Implement a "plugin" concept for validating extensions


javagl:
The validator currently does not support validation of extensions. There are _many_ technical caveats for that in the detail. But it might be possible to set up a basic "plugin" infrastructure for the validation of extensions, that could already have a great effort-to-benefit ratio.

One basic idea was to offer a point where it is possible to plug in implementations of the [`Validator` interface](https://github.com/CesiumGS/3d-tiles-validator/blob/125a6c667dedd4d9fb5374337094aff1117c83c3/src/validation/Validator.ts). In pseudocode, this could be
```
class S2ExtensionValidator implements Validator<any> {
  validateObject(input: any, context: ValidationContext): boolean {
    // TODO Validate the input
    return true;
  }
}
TilesetValidator.registerExtensionValidator(
    "3DTILES_bounding_volume_S2", new S2ExtensionValidator());
```
Whenever the validation encounters a [`3DTILES_bounding_volume_S2`](https://github.com/CesiumGS/3d-tiles/blob/f9ad4e1a4c753b91262c63121b3124f38ffe49ff/extensions/3DTILES_bounding_volume_S2/schema/boundingVolume.3DTILES_bounding_volume_S2.schema.json#L2) extension object, the given validator will be consulted for that validation.

A neat side effect of this is that it would be possible to perform the purely JSON-schema based validation _generically_ So creating a validator for a new extension, purely based on the schema, could be as easy as
```
const example= Magic.create("/example.schema.json");
TilesetValidator.registerExtensionValidator("EXAMPLE_extension", example);
```

But there are some details to sort out, and these are not entirely trivial. Extensions can sometimes not be validated "isolatedly". Some extensions affect the validity of the surrounding objects. One example is, exactly, the `3DTILES_bounding_volume_S2` extension: For a bounding volume like this
```
{
    box: undefined,
    region: undefined,
    sphere: undefined,
    extensions: {
    3DTILES_bounding_volume_S2: {  ... }
   }
}
```
one could validate the extension, **but** the bounding volume itself would still be considered to be invalid, because according to the specification, one of `box`, `region` and `sphere` must be defined.

A similar (but more generic) instance of this is when an extension extends the set of possible values for enums. For example, there might be an extension that defines a new `refine` value, and in addition to `ADD` and `REPLACE`, there is now a value like `MERGE` that is considered to be valid. 

[...]

PR #230, opened 2022-10-10
Drafts for extensions validation

javagl:
Addressing https://github.com/CesiumGS/3d-tiles-validator/issues/227 : This PR adds a mechanism for adding validators for extensions. Some technical details are described in the issue, and still have to be resolved here. 

This is only a basic draft, with support for the `3DTILES_bounding_volume_S2` extension: By default, a tileset with S2 bounding volumes would cause a validation issue, because the bounding volume neither has neither `box`, `region`, nor `sphere`. But with the extensions validation support, it would be possible to register a validator for S2 bounding volumes. The registration is currently *drafted* like this: 
```
const s2Validator = new BoundingVolumeS2Validator("");
const performDefaultValidation = false;
ExtensionsValidator.register("3DTILES_bounding_volume_S2", s2Validator, performDefaultValidation);
```
And afterwards, the original issue would no longer be reported. Instead, it would report validation issues for the S2 bounding volumes accordingly - like `S2_TOKEN_INVALID` or so...

Adding a _general_ support for extensions will require some refactorings of the current validator state. These refactorings should be "small" in terms of the amount of code that has to be changed, but they may affect the interfaces, and therefore several parts of the code. Further technical details will be added here soon.


javagl:
A short brain dump:

Some questions for the validation of extensions are not trivial to answer, and the S2 bounding volume extension (fortunately?) shows some of them.

The initial approach sketched above was:
- Register a new validator for the S2 extension
- When a bounding volume with S2 extension was found:
  - Skip the default validation of the bounding volume 
  - Validate the `3DTILES_bounding_volume_S2` object instead

This does not sensibly cover the case that there is a bounding volume that has the S2 bounding volume **and** a `box`, for example. In this case, the validation of the `box` should still be peformed. More specifically, in the case of S2, the pseudocode for the validation would be
```
if (!has(boundingVolume, "S2")) {
    validateAnyOf("box", "region", "sphere");
}
validateBox();
validateRegion();
validateSphere();
```
I.e. it should *only* skip the `anyOf` check, but still validate everything else. 

More generically, it's hard or impossible to define _exactly_ which aspect of the default validation should be omitted due to the presence of an extension. 

One "obvious" way of tackling this is:

- Register a validator for the S2 extension
- When a bounding volume with S2 extension is found:
  - Use the registered validator for the validation of the bounding volume. This registered validator would do the same as the default validator, _except_ for the `anyOf` check, and _additionally_ validate the `3DTILES_bounding_volume_S2` object

This still raises questions for the more general case, or for future extensions. For example, consider a future `3DTILES_bounding_volume_KDOP` extension (something that I'd like to see one day...). The process would then be

- Register a validator for the S2 extension
- Register a validator for the KDOP extension
- When a bounding volume with S2 *and* the KDOP extension is found:
  - This may cause duplicate issues, because the validation of the `box` is performed _twice_ as part of the default behavior

I'd like to avoid constraining what extensions are allowed to do - except for that extensions may not cause a _valid_ object to become _invalid_. But considering all possible ways of how an extension may cause an (otherwise) _invalid_ object to become _valid_ (and how this affects the overall validation process) is not entirely trivial. 

I'll update this PR soon, but wanted to sort my thoughts here.

javagl:
The current state contains a _basic_ implementation of a plugin concept for validating extensions, and various cleanups and generalizations for further working towards that goal.

There are two high-level points that are part of this PR:

---

It is possible to register a `Validator` for objects that contain certain extensions:
```
      const s2Validator = new BoundingVolumeS2Validator();
      const override = true;
      ExtendedObjectsValidators.register(
        "3DTILES_bounding_volume_S2",
        s2Validator,
        override
      );
```

(Note that this is a validator for the objects that _contain_ the extensions. and not for the extension objects themself - although the validation of these objects should be _part_ of the registered validator, of course).

Whenever the validator encounters an object that may contain extensions, it passes it to the `ExtendedObjectsValidators` class. This class will check whether the object contains an extension for which a new validator was registered, and applies this validator accordingly. When the `override` flag was `true`, calling this validator will cause the default validation to be skipped.

One example of such a validator is the [`BoundingVolumeS2Validator`](https://github.com/CesiumGS/3d-tiles-validator/blob/c95c2e1964a4b47d644c4148762f58623e0ad283/src/validation/extensions/BoundingVolumeS2Validator.ts#L13). This validator performs the basic validation of a _single_ `3DTILES_bounding_volume_S2` object. The way how this validator is integrated is shown in the [`BoundingVolumeValidator`](https://github.com/CesiumGS/3d-tiles-validator/blob/c95c2e1964a4b47d644c4148762f58623e0ad283/src/validation/BoundingVolumeValidator.ts#L60)

**Note**: This sort of validation has a scope that is still too narrow. There are some forms of extensions and associated validation that can _not_ be covered with this approach:

- The validation of a _single_ S2 bounding volume can not validate any form of _consistency_ of the S2 hierarchy. This part will require some form of generalization, which roughly means: Finding the right place to "hook in" something like a `Validator<TwoBoundingVolumes>` that will be called for parent+child and/or tile+content bounding volumes. Eventually, this could replace the [legacy `BoundingVolumeChecks`](https://github.com/CesiumGS/3d-tiles-validator/blob/c95c2e1964a4b47d644c4148762f58623e0ad283/src/validation/legacy/BoundingVolumeChecks.ts).
- There are certain extensions that do not require an explicit validation of one object. For example, an extension that allows additional content types, like `3DTILES_content_glTF`. Such an extension can affect the validity of a tileset as a whole, referring to places that are deeply hidden and unrelated to the point where the extension is declared.
 
---

The last point (namely, extensions that refer to content types) was the reason for another generalization that is applied in this PR. Until now, the [`ContentDataValidator`](https://github.com/CesiumGS/3d-tiles-validator/blob/2c89c2c6f0c2f0771e230e69e6203e8e8ca09103/src/validation/ContentDataValidator.ts#L109) contained a few hard-wired, specific checks for certain content data types. The necessity for generalizing this was obvious, and is therefore addressed here. It is not yet fully fleshed out, but should already be a reasonable step forward.

There now is a `ContentDataValidators` class (note the `s` - it might be renamed to `ContentDataValidatorRegistry` or so...). This class allows registering `Validator` instances that are used for different content types, based on different criteria. These critera may be the `magic` header, or the file extension, or vague things like ~"something is probably a tileset"...

An example for how the current set of content data validators is registered is shown in the [current state of the `registerDefaults` method](https://github.com/CesiumGS/3d-tiles-validator/blob/c95c2e1964a4b47d644c4148762f58623e0ad283/src/validation/ContentDataValidators.ts#L33).

The actual `ContentDataValidator` will then validate the content data: It will look up the `Validator` that is supposed to be used for the given content data, and apply it to the content data (as shown [here](https://github.com/CesiumGS/3d-tiles-validator/blob/c95c2e1964a4b47d644c4148762f58623e0ad283/src/validation/ContentDataValidator.ts#L118
)). 

**Note**: There are still a few questions that are not fully resolved. One question is about the connection between 1. an extension that allows a new content type and 2. the actual _validator_ for that new content type.

Roughly: 
- Where and how should an extension for `3DTILES_content_XYZ` be registered, alongside the `XYZValidator`?
- Related to https://github.com/CesiumGS/3d-tiles-validator/issues/231 : How to detect whether the extension is used? 
- (Important): How to encode the criterion for determining whether a certain content actually _does_ have the type "XYZ"? 

The latter is particularly difficult for JSON-based content types. **Iff** we already had 3D Tiles 2.0, then a valid 3D Tiles 2.0 file (with `asset.version = "2.0"`) would also be a valid glTF asset 😬 . So this has to involve some guesses. (The file extension may be a hint, but the extension is not required...).

javagl:
A basic infrastructure for the validation for extended objects (and extensions) is added with this PR. There are still some open questions about the _interface_ of how certain types of extension validators will be added or registered. But this will be summarized in a dedicated issue.

PR #236, opened 2022-10-18
Metadata validation

javagl:
This is a draft PR, tracking further steps for the validation of metadata. This primarily refers to _binary_ metadata, as it is stored in implicit subtrees. The validation of JSON-based metadata - aka `metadataEntity` objects - is already covered to a larger extent. (Certain parts of the validation of binary- and JSON-based metadata will have some overlap). 

Steps until now:

- Some (minor, internal) refactorings and cleanups for the existing classes, like moving certain validation steps into dedicated classes
- Moved all the metadata-related classes into a `validation/metadata` directory. The goal here would, roughly speaking, to make this part "as standalone as possible". Theoretically (and ideally), the same code could be used to validate the binary metadata from a `subtree` file **or** the binary metadata from `EXT_structural_metadata`. (This is not an _explicit_ goal, but the structural similarities should be kept in mind)
- Started to draft possible (unit) tests for the binary metadata validation. 
 
The latter is not entirely trivial: There are many constraints for binary metadata, for example, in the [descriptions in `propertyTable.propery`](https://github.com/CesiumGS/3d-tiles/blob/21aba4796c8281f33e3235493ae9040b70a50bc8/specification/schema/PropertyTable/propertyTable.property.schema.json#L11). For each constraint, there should be a dedicated test. The idea is, roughly speaking, to have one test for each of these constraints - like [propertyStringWithoutStringOffsets](https://github.com/javagl/3d-tiles-validator/blob/9ee592f2f8aded2f9bf907f7567116f38aad9030/specs/data/propertyTables/propertyStringWithoutStringOffsets.json), load and validate this file, and check whether the expected issues are generated ([example](https://github.com/javagl/3d-tiles-validator/blob/9ee592f2f8aded2f9bf907f7567116f38aad9030/specs/metadata/PropertyTablesSpec.ts#L27)). 

This is still _relatively_ simple on the JSON level, but becomes tricky when it really involves the data that is stored in the binary buffers: There must be buffer data with "invalid" contents (like `arrayOffsets` with negative or otherwise inconsistent entries). This has to be created for different configuations and types (like "Variable-length array with normalized INT32 MAT4 elements", or "Fixed-length enum arrays for a certain `enumValueType`"). It will likely not make sense to store this in _files_. Instead, it will probably (have to) be created as plain `Buffer` objects in the unit test itself, probably including an infrastructure that _generates_ these buffers from human-readable input. For example, something that can create the `values`/`arrayOffsets`/`stringOffsets` data for a variable-length string array that is given in JSON form.


javagl:
The [last commit](https://github.com/CesiumGS/3d-tiles-validator/pull/236/commits/f417ab932b7c17b53dd382643a06e050216d73be) contains a set of classes that are supposed to be used for the validation of binary metadata. These classes are, strictly speaking, not related to _validation_ per se. The main purposes of these classes are 1. _creating_ binary metadata as property tables (for validation tests), and 2. _accessing_ binary metadata (to validate its contents).

Originally, there was the idea to use the [CesiumJS `MetadataTester`](https://github.com/CesiumGS/cesium/blob/b4097de3b8d3d007ed38b3b6fb83717ab6de43ba/Specs/MetadataTester.js) for creating the metadata. But this class is heavily focussed on its use in unit tests, and created what could rather be referred to as "mock data". Additionally, there haven't been any structures for _accessing_ metadata (except for internal ones like [`MetadataTableProperty`](https://github.com/CesiumGS/cesium/blob/b4097de3b8d3d007ed38b3b6fb83717ab6de43ba/Source/Scene/MetadataTableProperty.js) and related classes). 

The current state contains a [`binaryMetadataDemos.ts`](https://github.com/CesiumGS/3d-tiles-validator/blob/f417ab932b7c17b53dd382643a06e050216d73be/src/binaryMetadataDemos.ts) that shows some examples of how the new classes are used. This does not look toooo spectacular, but ... that's the whole point: It is _supposed_ to look simple there.

The rough outline of the related classes:

- There is a `PropertyTableModels` class that contains some methods for creating `PropertyTableModel` instances. These are intended for first tests (and not yet properly documented). The only public method fow now receices a `ClassProperty` and a set of values, and creates a `PropertyTableModel` for that single property
- A `PropertyTableModel` internally creates a mapping from property IDs to `PropertyModel` instances. Each `PropertyModel` represents one column of the table. The exact implementation depends on the type of the property (e.g. `NumericPropertyModel`, `StringArrayPropertyModel` etc...)
- The `PropertyTableModel` offers one function to obtain a `MetadataEntityModel`. Specifically, this `MetadataEntityModel` represents one _row_ of the property table. It offers a method to obtain the values of properties, based on the property name. The values of these properties are then fetched from the respective `PropertyModel`

This state should still be considered to be a draft. Beyond the 'demo', it does not yet contain proper tests (although the demo is already _structured_ similar to unit tests). Methods for validating the input data (based on which these structures are supposed to be created) are still missing. Methods for creating more test cases have to be added - particularly, the test cases that are "intentionally broken in various ways", for the actual validation.

javagl:
I'll mark this as "Ready For Review" for now, because it can be reviewed. even though there are still aspects that are not validated, and the test coverage is not as large as it should be.

The aspects of the binary metadata (property table) validation that are currently covered:

- The general consistency on the JSON level, and buffer structure (things like checking that a `bufferView.buffer` is a valid index, and that the buffer views fit the buffer...)
- Some consistency checks for the `values`, `arrayOffsets` and `stringOffsets`: They have to be properly aligned, and their length must match for the `propertyTable.count` and the data type
- Some consistency checks for the `*Offsets`: They have to be increasing (i.e. non-decreasing)

One thing that is not validated yet is the actual values of the `values`, `arrayOffsets` and `stringOffsets`. These checks will include whether `arrayOffsets[i]` are actually valid offsets (i.e. range checks), and whether the values from the `values` match the declared `min` and `max` of the properties. Much of the infrastructure for that exists, and it should be implementable with reasonable effort, but some generalizations should be implemented as part of that - including an easy option to generate test cases that rely on `normalized` and `default/noData` values. (This is checked for the JSON representation, but not yet for the binary representation). 

---

Regarding the test approach and coverage: 

The general idea for now is to [create valid binary representations of property tables](https://github.com/javagl/3d-tiles-validator/blob/9154caaa48bf13718af325f222d4807759584a26/specs/metadata/BinaryPropertyTableValidationSpec.ts#L18), and then make sure that they are ["broken" in one way or the other](https://github.com/javagl/3d-tiles-validator/blob/9154caaa48bf13718af325f222d4807759584a26/specs/metadata/BinaryPropertyTableValidationSpec.ts#L167) for the tests. The tests in this file currently only cover the case of `INT16/SCALAR` and `INT16/variableLengthArray`. 

The cases that have to be added will probably be the ones that are currently part of the "positive" (non-validation) cases in [another test](https://github.com/javagl/3d-tiles-validator/blob/9154caaa48bf13718af325f222d4807759584a26/specs/metadata/PropertyTableModelsSpec.ts), namely:
```
  example_INT16_SCALAR
  example_variable_length_INT16_SCALAR_array
  example_fixed_length_INT16_SCALAR_array
  example_BOOLEAN
  example_variable_length_BOOLEAN_array
  example_fixed_length_BOOLEAN_array
  example_STRING
  example_variable_length_STRING_array
  example_fixed_length_STRING_array
  example_FLOAT32_VEC2
  example_variable_length_UINT32_VEC2_array
  example_fixed_length_UINT32_VEC2_array
```
That may look like a short list, but one has to keep in mind that there are _many_ ways in which a `fixed_length_UINT32_VEC2_array` could be "broken", and each of them will have to be represented as a test case for the respective validation issue.

lilleyse:
@javagl looks good, everything else you mentioned is fine to handle in future PRs.

PR #237, opened 2022-10-30
Metadata validaton - part 2

javagl:
The [Metadata validation PR](https://github.com/CesiumGS/3d-tiles-validator/pull/236) contained basic validation of _binary_ metdata from property tables. As mentioned in the [last comment there](https://github.com/CesiumGS/3d-tiles-validator/pull/236#issuecomment-1293703652), some aspecs of the binary data had not yet been validated, and some aspects of the existing validation had not yet been covered by spec-tests. This PR extends the current state in this regard. Further details will be added here via comments.

### DIFF
--- a/specs/jasmine.json
+++ b/specs/jasmine.json
@@ -1,6 +1,6 @@
 {
   "spec_dir": "specs",
-  "spec_files": ["**/metadata/PropertyTablesSpec.ts"],
+  "spec_files": ["**/*Spec.*s"],
   "helpers": [],
   "random": false
 }
--- a/specs/metadata/BinaryPropertyTableValidationSpec.ts
+++ b/specs/metadata/BinaryPropertyTableValidationSpec.ts
@@ -3,138 +3,12 @@ import { ResourceResolvers } from "../../src/io/ResourceResolvers";
 import { ValidationContext } from "../../src/validation/ValidationContext";
 import { BinaryPropertyTableValidator } from "../../src/validation/metadata/BinaryPropertyTableValidator";
 
-import { BinaryPropertyTable } from "../../src/binary/BinaryPropertyTable";
-import { PropertyTableModels } from "../../src/binary/PropertyTableModels";
+import { PropertyTableTestUtilities } from "./PropertyTableTestUtilities";
 
-import { Schema } from "../../src/structure/Metadata/Schema";
-import { ClassProperty } from "../../src/structure/Metadata/ClassProperty";
+// A flag to enable printing all `ValidationResult`
+// instances to the console during the tests.
+const debugLogResults = false;
 
-/**
- * Creates an unspecified valid default `BinaryPropertyTable`, containing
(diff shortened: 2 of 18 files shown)


javagl:
The current state adds an important fix for the creation of the underlying test data (fortunately, things like these are likely to be noticed as soon as the validation coverage increases), and a few refinements for the validations and tests. There now are dedicated checks (and appropriate error messages) for the case that `arrayOffsets` or `stringOffsets` contain values that require the byte length of the `values` to be different from what it actually is.  

Further tests for other data types will be added here soon.

javagl:
The [BinaryPropertyTableValidationSpec](https://github.com/CesiumGS/3d-tiles-validator/blob/8e2be3fcad094baf9f36cb151fef187e1ad2b98d/specs/metadata/BinaryPropertyTableValidationSpec.ts) now contain tests for all "categories" of property types, with "categories" referring to the property types
```
  example_INT16_SCALAR
  example_variable_length_INT16_SCALAR_array
  example_fixed_length_INT16_SCALAR_array
  example_BOOLEAN
  example_variable_length_BOOLEAN_array
  example_fixed_length_BOOLEAN_array
  example_STRING
  example_variable_length_STRING_array
  example_fixed_length_STRING_array
  example_FLOAT32_VEC2
  example_variable_length_UINT32_VEC2_array
  example_fixed_length_UINT32_VEC2_array
```
In these tests, I tried to cover all aspects, but to also avoid toooo much redundancy. For example, the check whether the `byteOffset` of a `values` buffer view has the proper alignment is not done for _each_ type repeatedly. So some of the (last) tests only consist of checking that a _valid_ input does _not_ cause any issue reports.

There have been a few minor internal fixes/"refactorings", which have mainly been the result of extending the test coverage for the binary property table validation. This included things like [creating elaborate error messages](https://github.com/CesiumGS/3d-tiles-validator/blob/8e2be3fcad094baf9f36cb151fef187e1ad2b98d/src/validation/metadata/BinaryPropertyTableValidator.ts#L673), which is cumbersome, but I hope that **iff** somebody encounters such a message, it will save a lot of debugging time (compared to a "simple" message like `"Byte length is invalid"` or so...).

The latest state also contains a first shot at the validation of the metadata _values_ - meaning the check that the actual values that appear in the metadat are in the range that is determined by the different `min`/`max` declarations. But some aspects of the values validation are not complete yet.

javagl:
The current state should be ready for review. 

Some aspects of the validation of _value ranges_ will be addressed in a follow-up PR, for example:

- Validating that the minimum/maximum of all values that are _present_ are actually the _same_ as the `min/max` that is declared in the `propertyTableProperty`
- Validating that the `min/max` from the `propertyTableProperty` is strictly in the range of the `min/max` of the `classProperty`
- Validating that the values of `ENUM` properties are only values that appear in the `enum.enumValue[i].values`

PR #238, opened 2022-11-02
Deeper binary metadata value validation

javagl:
A [previous PR](https://github.com/CesiumGS/3d-tiles-validator/pull/237) added basic validation of binary metadata values (as they are stored in property tables). But the validation of binary metadata values still has to be extended. 

The first one is the validation that binary values that are stored for `ENUM`-typed properties are actually values that appear as the `enum.values[i].value` of the respective enum definition in the schema. This is already covered in the current state of this PR:

The other one is the validation in view of the `min` and `max` that may be defined in a `ClassProperty` or a `PropertyTableProperty`. The `min/max` of the `ClassProperty` are supposed to be _global_ minimum/maximum value (meaning that there may be no value smaller than / greater than these). The `min/max` of the `PropertyTableProperty` are supposed to be the minimum/maximum values that _actually appear_ in the property table for the respective property.

The latter is ... "simple" from one perspective. (It involved some minor refactorings and cleanups, but nothing big). But it _does_ involve some efforts for a proper test coverage: The [current state of the validation specs](https://github.com/CesiumGS/3d-tiles-validator/blob/1300bbd78e6fdc569989c9c6bafb5f41d81db44d/specs/metadata/BinaryPropertyTableValuesValidationSpec.ts#L94) only cover relatively few cases:

- For `example_INT16_SCALAR`
  - Validate the values against the `classProperty.min`
  - Validate the values against the `classProperty.max`
  - Validate the values against the `propertyTableProperty.min`
  - Validate the values against the `propertyTableProperty.max`
  - Validate the `propertyTableProperty.min` against the values
  - Validate the `propertyTableProperty.max` against the values

This should (or has to?) be extended:

- All of the above for the case that a `classProperty.offset` is involved
- All of the above for the case that a `propertyTableProperty.offset` is involved
- All of the above for the case that a `classProperty.scale` is involved
- All of the above for the case that a `propertyTableProperty.scale` is involved
- ... 

More generally, there are 16 configurations of the `offset/scale` being present/absent in the `classProperty`/`propertyTableProperty`, the same for `normalized` being `true`/`false`, and the same for another type (like `VEC2/FLOAT32`), which will end up with ~"a few hundred" cases to be tested (only for the _numerical_ values, that is). I'll try to continue with generating some further tests here soon.


javagl:
I'll also think about whether some of these test can be "moved down one level". Roughly: To extensively check that things like the `offset/scale` are properly taken into account, so that this "dimension" can be left out of the tests, and just _assume_ that **iff** there is an `offset/scale`, the resulting `min/max` checks will yield the expected result (but it might be easy to overlook certain configurations in this case as well...)

javagl:
Part of this PR will be a validation that `min/max/offset/scale` are **not** given for _variable-length_ arrays. (This is mentioned indirectly in the spec text)

javagl:
The test cases of the binary property table value validation have been extended.

The cases that have been listed above for `example_INT16_SCALAR` have now been added [for `example_fixed_length_normalized_INT64_VEC2_array`](https://github.com/CesiumGS/3d-tiles-validator/blob/225871cb773da21aad3f06621948b1816e0155f1/specs/metadata/BinaryPropertyTableValuesValidationSpec.ts#L1213). The basic idea being here that this scenario covers many aspects that have not been covered in the first, "simple" case: 

- Arrays
- Structured types
- Normalized values
- An `INT64` component type

The alternative would be to "spread out" the tests for each of these aspects along each dimension, leading to a combinatorical explosion. Although such a more systematic approach for achieving a larger coverage here could be reasonable, it would raise the question of whether these tests should somehow be auto-generated. And I think that the coverage that can be achieved with this approach is reasonable as well: The focus of these tests is on whether invalid values _do_ cause the expected validation isses. The focus is _not_ whether the right values are obtained from the input. 

The latter should largely be covered by the [`MetadataEntityModelBasicSpec`](https://github.com/CesiumGS/3d-tiles-validator/blob/225871cb773da21aad3f06621948b1816e0155f1/specs/metadata/MetadataEntityModelBasicSpec.ts): It obtains the values for **all** property types, based on the `TilesetWithFullMetadata` sample, and compares them to the expected values. This, in turn, even though it has been "auto-generated" to a certain degree, still does not cover all aspects that are related to `offset/scale/default/noData`. But a _few_ tests for that are in the [`MetadataEntityModelSpec.ts`](https://github.com/CesiumGS/3d-tiles-validator/blob/225871cb773da21aad3f06621948b1816e0155f1/specs/metadata/MetadataEntityModelSpec.ts), and this could be extended separately.

In any case, I think that the current state is a reasonable step forward in terms for coverage for the binary metadata validation (and contains some bugfixes for the "JSON-level validation" as well), and is sufficiently consistent to mark this as ready for review.

PR #239, opened 2022-11-03
Archive validation support

javagl:
There are considerations for creating an "archive format" for 3D tiles (see [related issues in the `3d-tiles` repository](https://github.com/CesiumGS/3d-tiles/issues?q=is%3Aissue+is%3Aopen+label%3A%22temporary+-+packaging%22) for first thoughts about that)

As am experiment, this PR adds support for the validation of tilesets that are stored in [the 3TZ format](https://github.com/erikdahlstrom/3tz-specification/) as the tile content of other tilesets. This is **highly preliminary** for several reasons:

- The functionality for handling archives _itself_ was implemented only experimentally. Right now, it is a bunch of `.js` files in a subdirectory that have been copied here as-they-are. The medium-term goal might be to offer these files in a more "standalone" fashion, maybe as an own npm package, maybe in an own repository, maybe ported to TypeScript...
- The validation for now only supports 3TZ archives. But generalizing that to other archives (i.e. other implementations of the `TilesetArchive` interface) would be trivial, as soon as the general infrastructure for that is set up. 
- For now, it only validates 3TZ files that are _tile content_ (not standalone files). This is also easy to generalize.

There are some points in the validator that may require a slight refactoring/review for a proper integration here - nothing really "big", but with some caveats: 

Until now, the validation of tile content data was done by 1. downloading the content data as a `Buffer`, and 2. validating the contents of that `Buffer`. This does not make sense for an archive (particularly, because the experimental archive classes do not operate on `Buffer` objects, and not even on URIs, but on _file descriptors_ - ouch...). 

It's not obvious what could be a "smarter" approach here. Usually, the _type_ of content data is determined by the magic header of the binary data. So the data _has_ to be read into a `Buffer` (at least, partially...?). Or to put it that way: There is no reason why a file that is called `external.3tz` should not actually contain `B3DM` data. Relying on the file _extension_ is also brittle. 

One could consider to refactor the archive functions to work on `Buffer` data instead of _file descriptors_. But that may have drawbacks for _large_ archives (i.e. ones beyond what can reasonably be stored in a `Buffer`). 

The mechanisms for resolving data in the validator (surrounding the `ResourceResolver`) may therefore have to be extended. It was already relatively easy to create an `ArchiveResourceResolver implements ResourceResolver` that just loads resources from an archive (and that was the _goal_ of that interface in the first place!). But there are some places that are marked with `TODO` (mostly related to questions about relative URIs, or where to resolve which URI against which base URI to obtain the right URI for accessing the archive...), and it might be necessary to extend this in some way so that it does not immediately resolve the data as `Buffer` data, but is able to handle  _file descriptors_ (or URIs that are used as file descriptors) in some way.

And finally: The current state does not validate _the archive itself_. It is only used for _accessing_ the archive and validating the _contents_ of the archive (assuming that the archive is a valid 3TZ archive). Parts of the code for reading the 3TZ archive are based on [code in a branch that was originally intended to go into the validator](https://github.com/erikdahlstrom/3d-tiles-tools/blob/0b48a4287df0c0361961c7072f8eec49527df1ac/validator/lib/archive.js). But when the functionality for reading/writing archives was carved out, the actual _validation_ has been omitted, and will have to be added as dedicated validation code in something like a `TilesetArchiveValidator3TZ`...


javagl:
The question about "How to detect how to validate content data" is not entirely trivial.

Some aspects to keep in mind:

- Relying on the file extension is brittle. It may be applicable for some cases (like `.geojson`), but not a general solution
- Relying on the "magic bytes" of the data requires _accessing_ these bytes in the first place. When the data is a (potentially huge) archive, then it should not read a huge blob of data into a `Buffer`, only to examine the first few bytes. 
- When the data is an archive, then it will usually not make sense to read the data at all (e.g. when the archive is implemented based on a file descriptor)
- That general problem of JSON: It's easy to detect whether something is _probably_ JSON (namely, when it starts with `{` or `[`). But detecting whether it really _is_ JSON requires _parsing_ it, and even if it is, it is not clear whether it's glTF or a tileset.json

The last commit generalizes some of the "content type detection mechanisms" as an attempt to tackle this, but may still not be the final solution. 

- There is a `ContentData` object that serves as the "key" for lookups. It contains a URI and a `ResourceResolver`. 
- When the content data is determined by the file extension: Only the URI has to be examined.
- When the content data is determined by magic bytes: It will try to load _the first few bytes_ of the content data, and see whether these bytes are a known magic header. 
  - The `ResourceResolver` has therefore been extended with a `resolveDataPartial` function. The idea here is, _very_ roughly speaking, that it could later be possible to implement this with _range requests_. It should not download a 100GB file just to see that it does not contain any known magic header.
  When the magic header is known, the _full_ data is downloaded and passed to the respective validator.
- When the content data is some form of JSON: It will download the _full_ data, and see whether it is the expected JSON data

Right now, archives are _only_ detected by the file extension (`.3tz`). If they are supposed to be detected by magic bytes (i.e. the PKZip header), then this could be implemented accordingly. In any case, it should be possible to avoid loading the _whole_ archive with this approach. As soon as it detects that a URI points to an archive, it resolves the URI (into an absolute file path), and creates the archive using a file descriptor for that path.

javagl:
There are several questions about how the archive support will eventually be integrated. This refers to the fact that "the archive support" right now was only added by moving a set of JavaScript-files that contain basic implementations of the 3TZ functionality into a directory here. 

So this PR will probably remain in 'Draft' state until this is sorted out. 

But the current state should be consistent and functional insofar that it can validate a tileset where tile content refers to `.3tz` files, and it does validate the data that is _contained_ in this `3tz` file.

lilleyse:
Thanks @javagl

PR #242, opened 2022-11-09
Validation issue severities, filtering, and result format

javagl:
The main function of the validator is to create a `ValidationResult`. This is not much more than an array of `ValidationIssue` instances, where each issue describes "what has been wrong", differentiated by the severity of ERROR, WARNING, or INFO.

The validator adds _all_ encountered issues to the generated result. So there is no mechanism for _skipping the validation itself_ for a certain aspect. 

<sup>This would require smearing thousands of `if (options.shouldValidateXyz)` statements throughout the code, raising the unanswerable question of what `Xyz` should be. The only exception here might be an option to "skip content validation", but that's a detail for now.</sup>

So the idea was to prefer _collecting_ everything, and _filtering_ out unwanted issues later.

The input for the validator is inherently _hierarchical_: There may be a tileset, that refers to an external tileset, that refers to a B3DM, that contains glTF. This structure is reflected in the validation result by the fact that each `ValidationIssue` can have `causes`. The basic idea is that a single glTF tile content that causes 100 warnings should not cause 100 warnings in the result, but only _one_ issue - namely, a `CONTENT_VALIDATION_WARNING` that _summarizes_ these 100 warnings as its `causes`.

For example, the result of validating a tileset that is a `validTilesetWithExternalValidTilesetWithValidB3dmWithInvalidGlb.json` may look like this:

```
Validation result:
{
  "date": "2022-11-09T16:22:41.497Z",
  "numErrors": 1,
  "numWarnings": 0,
  "numInfos": 0,
  "issues": [
    {
      "type": "CONTENT_VALIDATION_ERROR",
      "path": "validTilesetWithValidB3dmWithInvalidGlb.json",
      "message": "Content validTilesetWithValidB3dmWithInvalidGlb.json caused validation errors",
      "severity": "ERROR",
      "causes": [
        {
          "type": "CONTENT_VALIDATION_ERROR",
          "path": "tiles/b3dm/validWithInvalidGlb.b3dm",
          "message": "Content tiles/b3dm/validWithInvalidGlb.b3dm caused validation errors",
          "severity": "ERROR",
          "causes": [
            {
              "type": "CONTENT_VALIDATION_ERROR",
              "path": "tiles/b3dm/validWithInvalidGlb.b3dm",
              "message": "Content tiles/b3dm/validWithInvalidGlb.b3dm caused validation errors",
              "severity": "ERROR",
              "causes": [
                {
                  "type": "CONTENT_VALIDATION_ERROR",
                  "message": "Declared length (13296) does not match GLB length (8944).",
                  "severity": "ERROR"
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

---

There are some degrees of freedom for the filtering operations that users may want. Some high-level examples:

- Filtering by a minimum severity level (e.g. "include/exclude everything that is >=WARNING")
- Filtering by explicit severity levels (e.g. "include/exclude everything that is ERROR _or_ INFO")
- Filtering by the `type` (e.g. "exclude VALUE_NOT_IN_RANGE and EXTENSION_NOT_SUPPORTED")
- Pruning away the `causes` for each issue (maybe only up to a certain "depth"?)
- ...

Some of these are equivalent or can be transformed into each other. Some of them could/should be offered in an "inclusive" form and in an "exclusive" form. 

For all of them, there is the question of whether they should be applied _only_ to the "top level", or to the `causes` of each issue recursively

---

Another question is whether content validation issues should be specific for the content type. Right now, errors in tile content always cause a `CONTENT_VALIDATION_ERROR` (with the respective `causes`). I think that it _could_ make sense to differentiate these, into something like `CONTENT_VALIDATION_ERROR_FOR_EXTERNAL_TILESET` and `CONTENT_VALIDATION_ERROR_FOR_TILE_DATA`. The reason is that certain filter operations may have to be applied to _tile data, but not to _external tilesets_. Maybe this should even be as specific as dedicated types `...FOR_GLB...`, `...FOR_B3DM...`, ... 

---

One thing to keep in mind is: It would be desirable to have these filter operations being described in a _serializable_ form - i.e. in a form that can be added to some `config.json` file. Roughly as in `filters: [ { includeSeverities: [ERROR, INFO] } ]` or `excludeTypes: [ VALUE_NOT_IN_RANGE ]`. 

---

I thought that it could make sense to offer these filter operations in a form like this:
```JavaScript
  const filtered = result
    .filter(byIncludedSeverities(ERROR, INFO))
    .filter(byExcludedTypes("EXAMPLE_ERROR_A", "EXAMPLE_INFO_C"));
```
<sup>(This is actual, working code, with some shortcuts for the containing class names)</sup>

There may be more complex operations that people might need - things like: "Filter out all warnings of type X that are the causes of warnings in B3DM content that is referred to by an external tileset". There is no need to support _arbitrary_ complexity, of course. But these operations (even if they are only offered in the most simple form in the first version) _will_ eventually have to be called by those who want to use the validator as a library, so I think that it does make sense to put some thought into how this part could evolve.


javagl:
**tl;dr**:

- A basic functionality for filtering `ValidationResult` instances has been added.
- Separated a `CONTENT_VALIDATION_...` issues and `EXTERNAL_TILESET_VALIDATION_...` issues
- Minor updates of main README and implementation notes

--- 

The filtering functionality for now is focussed on _filtering out_ validation issues that the client is not interested in. 

(There have been considerations to generalize this, and allow "transforming" issues (e.g. converting issues of one type into another - like converting an `ERROR` into a `WARNING`), but **iff** this is desired, it should be addressed as a dedicated task)

There currently are two "predefined" categories of filters:

- Filtering by _included severities_
- Filtering by _excluded types_

A basic usage example has been added as a small, standalone [`validationResultDemos.ts`](https://github.com/CesiumGS/3d-tiles-validator/blob/9f722db93f8151b415446887856437be286e7a22/src/validationResultDemos.ts#L102), and a short snippet for this has also been added to the main README:

```JavaScript
  const filtered = result
    .filter(ValidationIssueFilters.byIncludedSeverities(ValidationIssueSeverity.ERROR))
    .filter(ValidationIssueFilters.byExcludedTypes("EXTENSION_NOT_SUPPORTED"));
```

The main README now also includes the important **disclaimer**: The `validationIssue.type` is a plain string. And clients should be aware of the fact that (at least in the first releases), these strings _might_ change, even between minor releases. (The actual API for the library usage has not yet been decided upon, but it should be mentioned even for the case that clients are just examining the report files). 

An "unrelated" commit here: There are several external links pointing into the `tools` directory, and these caused a 404 until now. I added a `tools` subdirectory here, pointing possible visitors to the new `3d-tiles-tools` repository.

PR #243, opened 2022-11-22
Preparing an npm package of the validator

javagl:
Preparing to publish the validator as an npm package.  The state of this PR is experimental for now. Further details will be discussed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@ node_modules
 package-lock.json
 build
 .eslintcache
-
+*.tgz
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,16 @@
+/specs
+/demos
+/tools
+/node_modules
+.eslintcache
+.eslintignore
+.eslintrc.json
+.gitattributes
+.gitignore
+.npmignore
+.prettierrc.json
+.prettierignore
+tsconfig.json
+generateThirdParty.js
+ThirdParty.json
+*.tgz
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,5 @@
+
+
+Version 0.0.0 - 2022-MM-YY
+  
+  - Initial release
--- a/demos/binaryMetadataDemos.ts
+++ b/demos/binaryMetadataDemos.ts
@@ -1,7 +1,7 @@
-import { ClassProperty } from "./structure/Metadata/ClassProperty";
+import { ClassProperty } from "../src/structure/Metadata/ClassProperty";
 
-import { BinaryPropertyTables } from "./binary/BinaryPropertyTables";
-import { PropertyTableModel } from "./binary/PropertyTableModel";
+import { BinaryPropertyTables } from "../src/binary/BinaryPropertyTables";
+import { PropertyTableModel } from "../src/binary/PropertyTableModel";
 
 function runPropertyTableModelTest(
   name: string,
--- a/demos/implicitTilingDemos.ts
+++ b/demos/implicitTilingDemos.ts
@@ -1,17 +1,17 @@
 import path from "path";
-import { defined } from "./base/defined";
+import { defined } from "../src/base/defined";
 
-import { readJsonUnchecked } from "./base/readJsonUnchecked";
+import { readJsonUnchecked } from "../src/base/readJsonUnchecked";
 
-import { ResourceResolvers } from "./io/ResourceResolvers";
+import { ResourceResolvers } from "../src/io/ResourceResolvers";
 
-import { Quadtrees } from "./implicitTiling/Quadtrees";
-import { TemplateUris } from "./implicitTiling/TemplateUris";
-import { QuadtreeCoordinates } from "./implicitTiling/QuadtreeCoordinates";
-import { OctreeCoordinates } from "./implicitTiling/OctreeCoordinates";
+import { Quadtrees } from "../src/implicitTiling/Quadtrees";
+import { TemplateUris } from "../src/implicitTiling/TemplateUris";
+import { QuadtreeCoordinates } from "../src/implicitTiling/QuadtreeCoordinates";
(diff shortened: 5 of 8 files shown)


javagl:
The current state of this PR adds the basic steps that are required for the npm package.

---

In order to use the package as a command line application, it can be cloned, and the [command line options from `main.ts`](https://github.com/javagl/3d-tiles-validator/blob/0df860b5e4bfc565f389a1b63d25c6b786c124a2/src/main.ts#L10) can be used:

`npx ts-node src/main.ts --tilesetFile C:\Data\tileset.json`

---

For the usage of the package as a library, there is the question about _which_ part of the API should be made "public". 

Right now, there is an [`index.ts`](https://github.com/javagl/3d-tiles-validator/blob/0df860b5e4bfc565f389a1b63d25c6b786c124a2/src/index.ts) file that exports the elements that are supposed to be "public". The https://api-extractor.com/ tool seems to potentially be useful for tracking _what_ is actually exported as part of the API, and avoid accidental breaking changes. I applied that tool to the build output (see the notes about the [API Definition](https://github.com/javagl/3d-tiles-validator/blob/0df860b5e4bfc565f389a1b63d25c6b786c124a2/IMPLEMENTATION.md#api-definition) in the implementation notes). The output of this tool is shown here: [3d-tiles-validator.api.md](https://github.com/CesiumGS/3d-tiles-validator/files/10078341/3d-tiles-validator.api.md). This output suggests that the only the desired classes/interfaces are exported.  There are multiple methods in the `Validators` class, which are marked with warnings, but the only _non-internal_ method is `validateTilesetFile` - and that was the intention for now. The public API for now therefore consists only of

- The single `Validators.validateTilesetFile` method
- The `ValidationResult` with iths  `ValidationIssue` instances, and to which a `ValidationIssueFillter` may be applied

(Note that _everything_ that is exported there is still tagged as [`@beta`](https://tsdoc.org/pages/tags/beta/) - details about this may have to be discussed)

---

Installing the package (from the generated TAR file) into a fresh JavaScript project allows it to use the validator as follows:
```
const { Validators, ValidationIssueFilters } = require("3d-tiles-validator");

const resultPromise = Validators.validateTilesetFile(
  "tilesetGeometricErrorMissing.json"
);
resultPromise.then((result) => {
  console.log(result.serialize());

  const filteredResult = result.filter(
    ValidationIssueFilters.byExcludedTypes("PROPERTY_MISSING")
  );
  console.log(filteredResult.serialize());
});
```

javagl:
I added a stript for generating the API documentation with `api-documenter` (and added this in the implementation notes). From looking at the resulting documentation, I noticed some elements in the API that could be marked as `@internal`, and added this tag accordingly. 

(Note: I still have to familiarize myself with the visibility levels and implications of all this, but I think that the `@internal` tag only prevents the elements from _appearing_ in _some_ outputs. But it does not really _prevent_ them from being _used_, on the level of actual compile-time checks - there does not seem to be a concept of (real) "package private" elements in TS...)

---

One question that I'm not sure about is the `types` declaration in the `package.json`. The [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) says 

>  Set the types property to point to your bundled declaration file. For example: 
> ...
> "types": "./lib/main.d.ts"

A [stack overflow question](https://stackoverflow.com/q/48952313) suggests that this may hinder browsability in VSCode, but there are some technical details (and many 'but's, 'if's and 'should's,,,) in the question, comments, and answer that I do not (yet) fully undertand.

I pragmatically tried out both ways

- `"types": "src/index.ts"` and
- `"types": "build/index.d.ts"`

and installed the resulting TAR files in a consuming TS project. In **both** cases...

- running with `npx ts-node index.ts` seems to work
- running with `node dist/index.js` seems to work
- browsing to the definitions (with CTRL+click) in VS Code seems to work

So I updated it to the (suggested) `"types": "build/index.d.ts"` in the last commit. But I wonder whether that actually make a difference, or whether I missed something that may be relevant here. (I did not (yet?) find any specification about what the `types` property actually _does_ or how it is _used_ - but maybe that's intentional...)

IMPLEMENTATION.md
```diff
@@ -50,3 +50,55 @@ The `ValidationIssue` class and its types:
...
+
+The API definition is tracked with https://api-extractor.com
+
+- Install api-extractor:
+ 
+  `npm install -g @microsoft/api-extractor`
+
+- Install api-documenter:
+    
+  `npm install -g @microsoft/api-documenter`
```
lilleyse, line 63: Could these go in `package.json`?

IMPLEMENTATION.md
```diff
@@ -50,3 +50,55 @@ The `ValidationIssue` class and its types:
...
+
+- Run the TypeScript compiler to generate the build output:
+
+  `npm run build`
+
+- Invoke api-extractor:  
+
+  `api-extractor run --config api-extractor.jsonc --local --verbose`
```
lilleyse, line 71: This line fails for me on Windows / Git Bash if `etc` doesn't already exist:

```
Error: Unable to create the API report file. Please make sure the target folder exists:
C:\Users\slilley\Code\3d-tiles-validator\etc
```

IMPLEMENTATION.md
javagl: Done

IMPLEMENTATION.md
javagl: Done by updating the `npm run docs` script. But... when running it for the first time, it complains
```
Warning: The API report file was missing, so a new file was created. Please add this file to Git:
C:\3d-tiles-validator\etc\3d-tiles-validator.api.md
```
so it appears this file should be committed to git. This makes sense from the perspective of "tracking" the API and possible changes. But I find it a bit unfortunate. Clients might not be aware of the use of this file, and I hope that it does not cause them to assume that this is the API documentation or so...

Issue #7, opened 2016-08-31
Enable travis


pjcozzi:
https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/3#issuecomment-243894172

lilleyse:
Travis is enabled but broken at the moment. I attempted to fix it here: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/tree/travis, but didn't get much further than that.

shehzan10:
I think I have made a little more progress in my branch I did a while back. https://github.com/shehzan10/3d-tiles-tools/tree/travis and https://travis-ci.org/shehzan10/3d-tiles-tools/builds/224409856

The problem I had was with coveralls. Because the repo is setup as 3 distinct projects, coveralls always assumes that the current directory is the root. So all the 3 projects are going to generate coveralls that will then run into conflicts and the file links will not show up correctly. Here is the coveralls project link https://coveralls.io/builds/11181928.

Also has to xit one of the tests. Didn't look into it much.

mramato:
I disabled travis for the time being because having it work has not been a priority and it's just a bunch of noise.

javagl:
I consider travis to be one part of the somewhat broader question of deployment. Specificially: One part of the build process _might_ eventually be that the validator is deployed on a website with drag-and-drop support, similar to the glTF validator. The details and timeline for that have to be decided, but I wanted to mention it here to keep track of that.

PR #246, opened 2022-12-03
Travis setup

javagl:
This is supposed to set up travis. (This includes adding a `coverage` functionality with `nyc`. These could be independent PRs, but testing the coverage should be part of CI, so I hope it makes sense to _cover_ this here as well)

Trying this out locally would apparently require setting up a certain toolchain (e.g. as in https://stackoverflow.com/a/49019950 - with brittle, tedious, manual steps, copy-and-paste of job IDs and whatnot...) that makes me think "Well, just let it run on the server and see whether it works..." - but I'm open for either approach here...

If this "works", it fixes https://github.com/CesiumGS/3d-tiles-validator/issues/7 ...


javagl:
It's building. It's passing. I'll just mark this as "Ready For Review".

(An aside: The coverage does not look so great right now. This is to a large extent caused by the `tileFormats/(legacy)` part, and by some low-level helper classes that have been ported from CesiumJS and from which no all functions are actually used by the validator. Increasing the coverage e.g. by adding CMPT/PNTS/I3DM files in the `specs` tests, or removing the unused code paths could be considered as an ongoing task)

.gitignore
```diff
@@ -1,6 +1,8 @@
 /node_modules
 /build
 /dist
+/coverage
+/.nyc_output
```
lilleyse, line 5: These should also be added to `.npmignore`

lilleyse:
For new repos I think we should be using GitHub Actions. The setup should be pretty minimal.

javagl:
I converted travis into a GitHub action. This will only be run _after_ the corresponding file was merged (so it's even more difficult to test this beforehand), but from my first, basic experiments, the usage should be pretty straightforward (and if there's an issue with the YML, I'd fix that quickly)

lilleyse:
[ubuntu-latest](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md) comes preinstalled with Node 16 so it should be ok, but we'll see

lilleyse:
I pushed a couple commits directly to main to get it working: https://github.com/CesiumGS/3d-tiles-validator/commits/main

javagl:
Seems to work 🎉

PR #253, opened 2022-12-14
Packaging fixes

javagl:
Until now, after installing the validator from npm with 
`npm install 3d-tiles-validator`
it was necessary to start the build output _manually_, with something like
`node node_modules/3d-tiles-validator/build/main.js -t tileset.json`

This _should_ be fixed now, by adding a [`bin` entry in the `package.json`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin). This entry has to point to a JavaScript file that contains a very specific shabang for the Node environment. So unfortunately, this has to be a standalone file that is copied to the `build` directory before packaging.

But ... while doing that, I updated the `package.json` scripts to be a bit more fine grained, and to actually have a
`npm run package`
script that does everything that has to be done for generating the actual `3d-tiles-validator-X.Y.Z.tgz` for npm.

Opening this for an early review here.


IMPLEMENTATION.md
```diff
@@ -94,24 +94,21 @@ This API definition file is tracked with Git, so changes in this file should be
...
-  
-  `npm run docs` 
-
-  The documentation will be written as markdown files into `build/docs`.
 
-- Generate the tarball of the project:
+- Generate the tarball for the release:  
   
-  `npm pack` 
+  `npm run package`
```
lilleyse, line 100: I love how much simpler this is now

README.md
```diff
@@ -114,7 +122,27 @@ This will cause the validator to validate all JSON files in the specified direct
 
 
 
+## Developer Setup
 
+When the validator is not installed as a package from NPM, but supposed to be used directly in a cloned repository, then the command line usage is as follows:
+
+- Clone the repository into the current directory:
+  ```
+  git clone https://github.com/cesiumgs/3d-tiles-validator
```
lilleyse, line 131:
```suggestion
  git clone https://github.com/CesiumGS/3d-tiles-validator
```

package.json
```diff
@@ -51,14 +55,39 @@
     "typescript": "^4.8.3"
   },
   "scripts": {
+    "about:start": "echo 'Start the validator, as a TypeScript command line application'",
```
lilleyse, line 58: Is there a command to run to see the about messages?

javagl: This was one suggestion in https://stackoverflow.com/a/50878067. The messages can be displayed with something like
`npm run about:start`
but the primary goal for me was to _document_ these scripts in the `package.json` itself. 

I'm not really satisfied with that solution. Maybe I'll bite the bullet and move all these scripts into gulp. At least, _there_ it is possible to add comments, and do things like cleaning the output directory or copying a file without a bunch of dev-dependencies.

package.json
```diff
@@ -51,14 +55,39 @@
...
+    "about:docs-extract-api": "echo 'Extract the API definition file from the current build output.'",
+    "docs-extract-api": "api-extractor run --config api-extractor.jsonc --local --verbose",
+    "about:docs-generate-markdown": "echo 'Generate the API documentation based on the API definition file'",
+    "docs-generate-markdown": "api-documenter markdown -i build/api-extractor -o build/docs",
+    "about:docs-generate": "echo 'Generate the API documentation based on the current build output'",
+    "docs-generate": "npm run docs-prepare-directory && npm run docs-extract-api && npm run docs-generate-markdown",
+    "about:docs": "echo 'Perform a build and generate the API documentation based on the build output'",
+    "docs": "npm run build && npm run docs-generate",
+    "about:package-clean": "echo 'Clean the build output folder'",
+    "package-clean": "node -e \"require('fs-extra').emptyDir('build');\"",
```
lilleyse, line 85: `fs-extra` should be explicitly included in `devDependencies`. Also `emptyDirSync` is better to use here.

Is there a simpler way to do this? Maybe https://www.npmjs.com/package/del-cli?

package.json
lilleyse: The about messages do add a bit of clutter. I find the scripts mostly self explanatory, but up to you.

javagl: I think that it may be worthwhile to document the _preconditions_ of each script. For example, with
```
"about:docs-extract-api": "echo 'Extract the API definition file from the current build output.'",
"about:docs-generate-markdown": "echo 'Generate the API documentation based on the API definition file'",
```
it is clear that 
- the first one requires the build output to be up to date
- the second one requires the API definition file to be up to date

 (and maybe this could even be mode more explicit than it currently is)

But I agree that it looks cluttered. Here is what it could look like when all the `about` ones are grouped at the top: https://github.com/CesiumGS/3d-tiles-validator/blob/a2cd0f0a87072fd4bb8167e9f05bafe3a51cd723/package.json#L58

But all that is not perfect, and just some experiments. I'll try to find a better solution - preferably with a common way to handle tasks like delete/copy/mkdir that does _not_ involve one dependency per call.

lilleyse: This restructuring looks good to me, and having those three dependencies is not too bad.

javagl:
The `about:`-"scripts" are now put at the top of the `scripts` list. I still think that this is not perfect, but better than having _no_ documentation of the scripts at all. If this is moved to "proper" "build system" at some point, then these comments may be removed (or rather become the `/* Method comments */` for the analogous steps that might then, for example, appear in the `gulpfile.js`). 

I also added a few words about the build scripts in the `IMPLEMENTATION.md`, fixed one of the `demos`, and added a short `demos/README.md` to point potential readers to the demos that might currently be relevant for them.

Issue #256, opened 2022-12-20
b3dm samples - TilesetWithRequestVolume - city invalid?


bertt:
Hi, 

When validating the b3dm's in https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.0/TilesetWithRequestVolume/city, I noticed that ul.b3dm and ll.b3dm are not valid:

```
   "type": "CONTENT_VALIDATION_ERROR",
      "path": "ll.b3dm",
      "message": "ll.b3dm caused validation errors",
      "severity": "ERROR",
      "causes": [
        {
          "type": "BINARY_INVALID_ALIGNMENT",
          "path": "ll.b3dm",
          "message": "The byte length must be aligned to 8 bytes",
          "severity": "ERROR"
        }
      ]
    }
```

Looks like the glb byteLength is not aligned to an 8-byte boundary in these cases (ll.b3dm - 9700 bytes, ul.b3dm - 9684 bytes). 

Question: Is the validator correct in these cases? If so how to fix the b3dm when the glb byteLength is not aligned to an 8-byte boundary? 

I tried to fix this validator error by adding padding but then I got an other warning:

```
       "causes": [
            {
              "type": "CONTENT_VALIDATION_WARNING",
              "message": "Extra data after the end of GLB stream.",
              "severity": "WARNING"
            }
          ]
```

It would be helpful to have valid b3dm sample files.

javagl:
Just a short ping for now:

The validator is correct here. There are some data sets that cause validation issues which have only been detected with the new validator. These data sets include unit test- and sample data in CesiumJS, as well as some of the older sample models in the 3d-tiles-samples repository. (We recently started addressing these issues, e.g. by fixing the spec data via https://github.com/CesiumGS/cesium/pull/10852, but we plan to also apply similar fixes to all other tilesets that are publicly available). 

For the `ll.b3dm` in particular: This B3DM file was originally intended to be "just some dummy object". Unfortunately, it appears in _many_ samples and tests (and ... sometimes the file sizes of these "instances" are different - I think none of them is really "valid" in terms of the alignment requirements). 

I just did some experiments, with some preliminary, _in-flight_ results shown in [PaddingIssues-2023-01-03.zip](https://github.com/CesiumGS/3d-tiles-validator/files/10339999/PaddingIssues-2023-01-03.zip). It contains 4 versions of a dummy tileset that refers to the `ll.b3dm`:

[...]

PR #258, opened 2023-01-04
Take padding bytes in GLB data into account

javagl:
Corresponding to the proposal at https://github.com/CesiumGS/3d-tiles/pull/729


javagl:
The current state here is a draft for taking the possible padding bytes into account. On the pedantic level, it is **not** (yet) a reasonable implementation: The GLB data is supposed to be sent to the glTF validator. But in order to _extract_ the (real, padding-free) GLB data, one has to access the GLB header `length` field, which already makes the assumption that the given data **is** GLB data with a valid header. Further checks for the validity of the GLB header have to be inserted before accessing the `length` field here.

javagl:
When it is merged, this fixes https://github.com/CesiumGS/3d-tiles-validator/issues/256

PR #265, opened 2023-01-23
Package file validation

javagl:
There already was some experimental support for validating 3D Tiles archives (see [README.md in archives directory](https://github.com/CesiumGS/3d-tiles-validator/blob/158ab571bb155b74f3417118835608527d992d0c/src/archives/README.md)). This validation was only applied to cases where the tile _content_ was an archive, specifically (only) a 3TZ archive. It was not possible to validate a "standalone" archive. 

This PR extends the functionality so that standalone archive files can also be validated. This is mainly accomplished by making the functionality that already existed in the `TilesetArchiveValidator` more accessible. 

One point that has to be sorted out: 

The archive/package file can just be passed as the `--tilesetFile` at the command line. Alternatively, there could be a new command line argument, namely `--tilesetPackage`. 
(A similar question is whether the same difference should appear on the API level, as in `Validators.validateTilesetFile` vs. `Validators.validateTilesetPackage`)

I'm slightly leaning towards the latter: A "tileset file" should be a `tileset.json`, and an archive/package is simply not a "tileset file", but an entirely different thing. On the other hand, one could argue that it is convenient to only have the `tilesetFile` CLI argument for both cases...


lilleyse:
> On the other hand, one could argue that it is convenient to only have the `tilesetFile` CLI argument for both cases...

That's my preference. One less option to remember.


Also make sure to update CHANGES.md

PR #271, opened 2023-04-20
Unzip package entries if they are zipped

javagl:
The entries in packages like `.3dtiles` and `.3tz` files may be zipped _individually_. This was not yet anticipated by the validator. Now, resources that are accessed from these packages will be unzipped if they are zipped. 

Corner cases: What should happen if someone puts a `myData.tar.gz` file into such a package, and does it **not** want to be uncompressed transparently? (Related: How can this be made clearer in the spec at https://github.com/CesiumGS/3d-tiles/issues/727 ?)


javagl:
This is now part of https://github.com/CesiumGS/3d-tiles-validator/pull/270

Issue #276, opened 2023-06-11
Validator fails to resolve GLB from I3DM with URI with spaces


javagl:
When an I3DM refers to GLB data with a URI (instead of "inlining" the GLB data in the I3DM body), and the URI contains trailing spaces, then the validator fails to resolve this data.

According to the ["Padding" section in the specification](https://github.com/CesiumGS/3d-tiles/tree/62832c3d8294e2c530ac928f6ef4ccc62f4d176f/specification/TileFormats/Instanced3DModel#padding) :  

> Otherwise, if the glTF field is a UTF-8 string, it shall be padded with trailing Space characters (`0x20`) to satisfy alignment requirements of the tile, which shall be removed at runtime before requesting the glTF asset.

But right now, the validator only handles the case that the data is padded with `0x00` characters (in the [`I3dmValidator`](https://github.com/CesiumGS/3d-tiles-validator/blob/068f8bc88af1ddfabf5f2526a1a857f21f8df625/src/tileFormats/I3dmValidator.ts#L270). But it also has to handle trailing spaces. 

This was brought up at https://community.cesium.com/t/i3dm-uri-padding/24696

PR #277, opened 2023-06-11
Handle URI with spaces in I3DM

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-validator/issues/276 : 

Until now, the validator handled the case that the URI inside an I3DM was padded with `0x00` (zero) characters. But the specification says

> Otherwise, if the glTF field is a UTF-8 string, it shall be padded with trailing Space characters (`0x20`) to satisfy alignment requirements of the tile, which shall be removed at runtime before requesting the glTF asset.

So the validator has to handle the case that the URI is padded with `0x20` (space) characters. 

One question is still open: Should the validator consider it as an **ERROR** when the URI (i.e. the payload) is padded with `0x00` characters instead of spaces? 

Right now, it silently allows that. But sticking to the specification text, I'd say "Yes, this should be an error!". 

If this is confirmed, I'd update this PR to treat this as an error (and this aspect has to be reviewed in the `3d-tiles-tools` in the context of https://github.com/CesiumGS/3d-tiles-tools/issues/45 )


javagl:
(Note: Depending on the last decision here, the change log still has to be updated)

javagl:
Treating trailing `0x00` bytes as an ERROR would potentially cause existing (formerly "valid") files to become invalid. This may be done in the future, but only after carefully evaluating the implications. 

For now, the change from this PR is:

- Any `0x00` (zero) bytes in the URI buffer cause a WARNING
- Trailing `0x20` (space) bytes are properly trimmed

Issue #283, opened 2023-10-10
Allow validation of tile content files


javagl:
Right now, the validator uses a tileset JSON file as the entry point for the validation, given as `--tilesetFile` at the command line (or a whole bunch of files, via the `--tilesetsDirectory`)

It should also be possible to apply the validator to tile content files like B3DM, PNTS, or GLB _directly_. Until now, there was not sooo much demand for that: Large parts of the validation of the "legacy" tile formats also used the "legacy" validation code, and the corresponding reports might be a bit short. For the validation of GLB, there is the glTF-Validator, which is unparalleled (and actually _used_ by the 3d-tiles-validator for glTF content). 

But with the [glTF metadata extension validation PR](https://github.com/CesiumGS/3d-tiles-validator/pull/280) soon being merged, there could be a demand for validating glTF assets that contain the `EXT_mesh_features` and `EXT_structural_metadata` extensions. 
This would _also_ (mainly) consist of calling the glTF-Validator, but do the additional validation for the extensions, and include the results in the report.

This could be done pragmatically, with a CLI argument like `--tileContentFile example.glb`.

PR #285, opened 2023-10-12
Add support for validating single content files

javagl:
Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/283 : 

This adds a `--tileContentFile` command line argument for validating single content files (that are not part of a tileset JSON). For example,
`npx ts-node .\src\main.ts --tileContentFile ./specs/data/tilesets/tiles/glTF/TriangleWithWarnings/TriangleWithWarnings.gltf`
will generate a report including the warnings from the glTF validator.

The function will just call the "content validation" infrastructure that is otherwise used for the `content.uri` fields of the tileset JSON. So it will support all "known" content types (and create a `CONTENT_VALIDATION_WARNING` for content types that are not supported). 

<sup>The fact that this function can _also_ be applied to tileset JSON files is... 😬 ... well, a result of the fact that tileset JSON files also _are_ tile content.</sup>

We could consider to build a larger infrastructure around this. For example: Right now, there's also the `tilesetsDirectory` option that validates _all_ tilesets in a directory. And we could consider adding a `tileContentsDirectory` to validate all _files_ in a directory. But ... this could easily be misused. For example: When running this on a directory that contains a `tileset.json` that refers to all files, then all files would be validated twice, once as "input files", and once as "contents of the `tileset.json`". Given the primary goal of this function (to validate glTFs with metadata extensions, after https://github.com/CesiumGS/3d-tiles-validator/pull/280 ), the scope could/should be limited to _single_ files for now.

To do (in the current state): Update the README.md with documentation for the new command line option.


javagl:
This has been updated to build on https://github.com/CesiumGS/3d-tiles-validator/pull/284 (even though it is largely independent of the actual validation functionality, and only affects the CLI), and the README.md has been updated with an example of the new CLI command.

PR #291, opened 2023-11-08
Limit number of glTF validation issues

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-validator/issues/290

glTF tile content is validated with the glTF-Validator. When validating a glTF asset that contains _many_ errors, this can lead to an out-of-memory error. (For example, the GLB in the linked issue would generate roughly 6 million issues...). 

The glTF validator offers an option to limit the number of issues that are reported. It's hard to pick a "universally sensible" value for that. 

(One critical corner case could be that when the limit is set to `x`, then there may be an asset that generates `x` warnings or infos, and the `x+1`th issue would be an 'error', leading to the asset counting as 'valid' even though it isn't...)

This PR just adds a fixed `maxIssues: 1000` for now, which prevents the issue. Questions beyond that are: 

- How many issues should be the default? 
- Should this be configurable from the 3d-tiles-validator side (via validation options in a config file)? 
  - (If yes, some thought would have to go into that. It could be a `maxContentIssues` for _all_ types of content, but only the glTF Validator currently supports such a 'limit'...)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,8 @@
 
+Version ?.?.? - yyyy-mm-dd
+
+- The maximum number of issues that are reported for a single glTF asset is now limited (via [#291](https://github.com/CesiumGS/3d-tiles-validator/pull/291)).
+
 Version 0.5.0 - 2023-10-24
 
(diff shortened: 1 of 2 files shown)


javagl:
There is now an upper limit for the number of issues that are reported by the glTF-Validator. 

This limit is 1000 for now, but this not specified. The point is: When a glTF asset generates 1000 issues, then this is a _single_ issue, namely, 'the glTF asset being 'bad''. 

If necessary, that limit could be made configurable (via the validator config JSON file). But it's unlikely that an end-user would like to change this value, say, from 1000 to 1200. If necessary, it can always be made configurable later.

PR #293, opened 2023-12-26
Only issue a warning for channel mismatches

javagl:
Until now, the validator made the assumption that image channels have 8 bits. This means that when someone tried to represent, for example, a 16-bit value in a property texture with a _single_ channel, then the validator generated an `ERROR` with type `TEXTURE_CHANNELS_OUT_OF_RANGE`.

There still are some open questions about the handling of non-8-bit channels in images. Some thoughts and discussion are tracked in https://github.com/CesiumGS/3d-tiles/issues/748 . But for now, the validator should not claim that assets that contain images with 16-bit channels are "invalid". So this PR changes the former `ERROR` in these cases into a `WARNING` with the type `TEXTURE_CHANNELS_SIZE_MISMATCH`. 

The issue will therefore be reported as 
```
        {
          "type": "TEXTURE_CHANNELS_SIZE_MISMATCH",
          "path": "example.glb/propertyTextures/0/properties/exampleProperty",
          "message": "The property 'exampleProperty' has the component type UINT16, with a size of 2 bytes, and the type SCALAR with 1 components, resulting in 2 bytes per element, but the number of channels in the property texture property was 1",
          "severity": "WARNING"
        }
```
but the asset will still be considered to be valid.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 Version ?.?.? - yyyy-mm-dd
 
 - The maximum number of issues that are reported for a single glTF asset is now limited (via [#291](https://github.com/CesiumGS/3d-tiles-validator/pull/291)).
+- When the number of bytes that are required for a certain feature ID or property texture property did not match the number of `channels`, then the validator reported this as an `ERROR`, with the type `TEXTURE_CHANNELS_OUT_OF_RANGE`. This could cause errors to be reported for the case of 16-bit channels in textures, where these numbers do not have to match. Now, these cases are only reported as a `WARNING`, of the type `TEXTURE_CHANNELS_SIZE_MISMATCH`.
 
 Version 0.5.0 - 2023-10-24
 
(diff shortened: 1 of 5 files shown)


lilleyse:
Are there any changes required for `EXT_mesh_features` validation?

javagl:
The main aspect of this change is specific for the `EXT_structural_metadata` property textures. Namely: It checked whether the number of bytes of the data type of the _property_ matched the number of channels. 

For the `EXT_mesh_features`, there is no such constraint. It does only validat the basic structure of the `channels` (e.g. [here](https://github.com/CesiumGS/3d-tiles-validator/blob/5656f67378623dfb47e281aee9ad09284016c737/src/validation/gltfExtensions/ExtMeshFeaturesValidator.ts#L410). However, there are some open questions that are tracked in the issue at https://github.com/CesiumGS/3d-tiles/issues/748 . For example, the validator currently makes the [_assumption_ that the channels have 8 bits](https://github.com/CesiumGS/3d-tiles-validator/blob/5656f67378623dfb47e281aee9ad09284016c737/src/validation/gltfExtensions/ImageDataReader.ts#L74), and we still have to sort out how to handle cases where this assumption is wrong...

Issue #231, opened 2022-10-12
Clarify extension usage validation


javagl:
Currently, the use of extensions is validated in (what I first thought to be) a straightforward way. One aspect of that validation is:

- When there is an extension found during the traversal (for example, a `boundingVolume.extensions["3DTILES_bounding_volume_S2"]`), then the `"3DTILES_bounding_volume_S2"` is stored as a "found" extension in the validation context
- At the end of the validation, the "found" extensions are compared to the `extensionsUsed` that have been declared. When an extension was "found" but not declared as "used", a warning is created. When an extension was declared as "used", but not "found", a warning is created.

The problem (which is obvious in hindsight): Certain extensions are never "found" explicitly. For example, `3DTILES_content_gltf`. It never appears as an object anywhere, but only _allows_ glTF/GLB files to be used as tile content.

This can be solved, to some extent, by keeping track of the _actual_ content data types that have been encountered. (There are corner cases where this becomes less trivial. For example, when the content data validation is disabled via a command line flag).

A related question is the _version dependency_. A specific example is that `3DTILES_content_gltf` should probably not be used in a `1.1` tileset, and using it should cause a warning. The more general question is: How to encode the mix-and-match between "tileset versions" and "allowed extensions". There may be extensions that are valid only in `1.0`, other may be valid only in `1.1`, and others may be valid in both...

PR #305, opened 2024-03-25
Fix external tilesets extensions validation

javagl:
There are some open questions and tricky corner cases for the validation of `extensionsUsed` in a tileset. Some of this is tracked in https://github.com/CesiumGS/3d-tiles-validator/issues/231 . But in the current state, this part of the validation contained a clear-cut bug for the validation of the `extensionsUsed` when it came to _external_ tilesets. Specifically, in this configuration:

- A `tileset.json` refers to an `external.json`
- The `tileset.json` contains an `VENDOR_example_extension`
- The `tileset.json` declares the `VENDOR_example_extension` in its `extensionsUsed`

In this case, the validator erroneously reported an `EXTERNAL_TILESET_VALIDATION_ERROR`, stating that

>The extension 'VENDOR_example_extension' was found, but not declared in 'extensionsUsed'

This is a bug. The extension does not have to be declared as `extensionsUsed` in the _external_ tileset when it is only used in the _main_ tileset.

This PR fixes this issue:

When the validation of an external tileset starts, then the 'extensionsFound' (which track the extensions that are _actually contained_ in that tileset) will start with the _empty set_. It is then filled with the extensions that are found, and checked for consistency with the `extensionsUsed` only _within_ that external tileset. 
Afterwards, the 'extensionsFound' from the external tileset will be added to the 'extensionsFound' of main tileset (because when they are found in an external one, they _also_ have to be declared in the `extensionsUsed` of the main one). 

Further test cases have been added, in order to have a reasonable test coverage for the possible cases. These are summarized in the [`externalTilesetExtensions/README.md`](https://github.com/CesiumGS/3d-tiles-validator/blob/69384a695dc7c8cff47f70d35bd0b6961575a7c5/specs/data/tilesets/externalTilesetExtensions/README.md) file.

### DIFF
--- a/specs/ExternalTilesetExtensionsValidationSpec.ts
+++ b/specs/ExternalTilesetExtensionsValidationSpec.ts
@@ -1,0 +1,80 @@
+import { Validators } from "../src/validation/Validators";
+
+// Note: These test cases use a dummy extension called
+// VENDOR_example_extension
+// So they will always create at least a WARNING about
+// this not being supported
+
+describe("External tileset extensions validation", function () {
+  it("detects no errors in declaredInBothContainedInExternal", async function () {
+    const result = await Validators.validateTilesetFile(
+      "specs/data/tilesets/externalTilesetExtensions/declaredInBothContainedInExternal/tileset.json"
+    );
+    expect(result.length).toEqual(1);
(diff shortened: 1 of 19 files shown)


lilleyse:
Thanks @javagl

Issue #226, opened 2022-10-04
Review the unit test approach


javagl:
For each aspect that is tested by the validator, I created a tileset JSON file that was supposed to cause _that exact issue_. The result is a set of serveral hundred JSON files in the `specs/data` directory, for tilesets, metadata, and subtrees. 

The current approach of most of the unit tests (e.g. in [`TilesetValidatorSpec`](https://github.com/CesiumGS/3d-tiles-validator/blob/a4eac895836b86347ab5fc310f534c809d5a4829/specs/TilesetValidationSpec.ts) ) is to read each of these files, and check whether the validator generates the expected issue. 

For some tests, it would probably make more sense to _not_ read the data from a file, but to perform the validation call on an object that is directly declared in the test code itself. But having the set of files (that can easily be checked individually, at the command line) turned out to be useful occasionally, even though it may be considered to be more of an "integration test" than a "unit test". 

I also considered the option to summarize these test cases in their own JSON file, roughly like in
```
[
  {
    "inputFile": "boundingVolumeRegionArrayInvalidElementType.json",
    "description": "One element of the region array is not a number",
    "expectedIssues": [ "ARRAY_ELEMENT_TYPE_MISMATCH" ]
   }
   ...
]
```
This could simplify testing in some ways, and allow _documenting_ the test via the `description`. But it might turn out to be an unnecessary maintenance burden.

If somebody has preferences, thoughts, or alternative approaches, feel free to discuss them here.

javagl:
This question came up, more prominently, in https://github.com/CesiumGS/3d-tiles-validator/pull/319

There are a few points to consider.

---

First of all, the _number_ of spec files is huge. This was not so much a problem for the 3D Tiles validation. But the `NGA_gpm` schema has a different structure: It is nested _very_ deeply. Note that in 3D Tiles, the JSON can be arbitrarily deep - but that's not the point: All the validation still happens on properties of the `tileset` and `tile` and its properties. In `NGA_gpm`, there literally is a property 
`root.extensions["NGA_gpm"].masterRecord.collectionRecordList[0].sensorRecords[0].collectionUnitRecords[0].lsrAxisUnitVectors[0]`
which is supposed to be a 3D unit vector. In order to have a complete (valid) tileset where _only this property_ is wrong, there have to be files where ...
- this unit vector has only 2 components
- this unit vector has 4 components
- this unit vector contains a `"NOT_A_NUMBER"` entry
[...]

PR #319, opened 2024-11-09
Add `NGA_gpm` validation

javagl:
This builds upon the state from https://github.com/CesiumGS/3d-tiles-validator/pull/316 - that PR also provides some general context.

This adds support for the validation of the `NGA_gpm` extension, which is the 3D Tiles extension for GPM (corresponding to `NGA_gpm_local`, which is the glTF extension).

I'm opening this as a draft for now:

- The other PR should be merged first
- There are some open questions about the GPM specification/schema that fell out of the implementation here, and that are still being discussed internally
- There are no specs here yet

The last point is _strongly_ related to https://github.com/CesiumGS/3d-tiles-validator/issues/226 : I locally *do* have the necessary specs. But _creating_ them was an odd mix of semi-automation and lots of (lots of) manual work, tweaking the files to be "broken" in one way or another, carefully matching the different ways of being "broken" with the spec and the schema. It still has to be decided whether another approach should be taken here.

A draft PR that shows the specs and the corresponding files is in https://github.com/CesiumGS/3d-tiles-validator/pull/320 

---

(Unrelated: I snuck in the (EDIT: _specs_ for the) validation of the `3DTILES_bounding_volume_S2` extension here. The files for that already did exist. But the actual `...Specs.ts` had been missing. This was only a minor fix that went along with the minor restructuring of the specs directory that I did for this PR)


javagl:
Merged in the specs that had been shown as a preview in https://github.com/CesiumGS/3d-tiles-validator/pull/320 

Added validation of aspects that are not covered by the JSON schema itself:

- Validate that `unitVector` has unit length (epsilon=0.00001)
- Validate that `lsrAxisUnitVectors` form an orthonormal basis (dot product epsilon=0.0005)
- Validate that `referenceDateTime` is a valid ISO8601 string
- Validate that the lengths of "upper triangular matrices" are in fact triangular numbers
- Validate that the `ppeManifest[i].source` values are unique

Added validation of aspects that are _currently_ not covered by the JSON schema. This refers to aspects that are not covered by the current ("preview") version of the JSON schema, but - as confirmed via internal discussion - should become part of the schema, and therefore also be validated: 

- The `referenceSystem` may not contain an `epoch` when it has a `definition`
- The `modelCoordSystem` should not contain properties that it does not need, depending on its `mcsType`
- The `anchorPointMetadata.contentIndex` should not be negative
- In `NGA_gpm_local`:
  - The `covarianceDirectUpperTriangle` should have at least a length of 1
  - The `anchorPointsIndirect` and `anchorPointsDirect` should have a length of at least 1
  - The `covarUpperTriangle` should have a length of at most 6 (i.e. it should have a length of _exactly_ 6)

javagl:
As mentioned above, one aspect of the validation is

> Validate that `lsrAxisUnitVectors` form an orthonormal basis (dot product epsilon=0.0005)

We now encountered data sets that are _likely_ supposed to be valid, and where these dot products had been in the range of `0.0088...`. So we preliminarily changed that epsilon to `0.01`. This sounds like a lot. It may have to be adjusted in the future.

PR #320, opened 2024-11-09
Add specs for GPM validation

javagl:
This is a draft PR, built on top of https://github.com/CesiumGS/3d-tiles-validator/pull/319 , to illustrate the effect of adding the specs. It still has to be decided whether this is the approach that should be used here.


javagl:
This has been merged into the (pending) PR https://github.com/CesiumGS/3d-tiles-validator/pull/319

Issue #326, opened 2024-12-05
User-defined semantics


javagl:
(This was originally considered to become part of https://github.com/CesiumGS/3d-tiles-validator/pull/325 , but may warrant its own issue:) 

At some point, we'd like to address https://github.com/CesiumGS/3d-tiles/issues/643 and https://github.com/CesiumGS/3d-tiles/issues/574 . And I think that one important aspect of that is to establish a mechanism for how these new semantics are brought into the validator. The first issue already suggests that new semantics could (nearly) be defined with a pure, plain 3D Tiles Metadata schema, using the 'semantics' as _property names_ in that schema. And that's basically what is already [done in the validator](https://github.com/CesiumGS/3d-tiles-validator/blob/ccae67559a82a39ce21eb6dcde9274a85a0edc8d/src/validation/metadata/ClassPropertySemanticsValidator.ts#L254). There is the minor caveat that the `componentType` in this schema can also be a RegEx like `FLOAT(32|64)`, to match the different options of component types. But I'd suggest to simply disallow this for future semantics definitions.

Given that the validation of semantics then boils down to loading a plain, pure, 3D Tiles Metadata schema and comparing its properties to other properties, the generalization here would be:

- Move the "matching schema" that is currently used here into a file, like `CesiumMetadataSemantics-0.0.1.json`
- Load this file and use it for the validator
- (Eventually:) Allow users to define their own "matching schema files", and also use them for the validation

The last point means that users could define their semantics in a schema, like
```
{
  id: "CompanyNameMetadataSemantics-0.0.0",
  classes: {
    CompanyNameSemantics: {
      properties: {
        COMPANY_NAME_SEMANTIC_A : {
          description: "What this semantic is...",
          type: "STRING",
          array: true
        }
      }
    }
  }
}
```
and pass in a command line argument like
`... -semantics ["CompanyNameMetadataSemantics-0.0.0.json"]`
to take these semantics into account during the validation.

javagl:
This was addressed with https://github.com/CesiumGS/3d-tiles-validator/pull/329 

Further changes can be tracked in deicated issues.

PR #329, opened 2024-12-17
Custom semantic schema registration

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-validator/issues/326 :  

The approach here is:

- The `ValidationOptions` contain `semanticSchemaFileNames: string[]`
- The `ValidationContext` contains `semanticMatchingSchemas: Set<any>` 
  - (This should ideally be a set of `Schema` objects, but cannot be due to the component type regex...)
- When a new `ValidationContext` is created, then it is filled with the default Cesium Metadata 'matching schema' and with the schemas that are read from the files that are defined in the validation options
- The `ClassPropertySemanticsValidator` examines the 'matching schemas' that are found in the validation context, and performs the semantic validation against them

Whether or not it should really be possible to also define the "matching schemas" at the command line has to be decided. I'd lean towards not supporting that, given the effort-to-benefit ratio: The "options file" approach seems more convenient. Specifically, there can be an options file like
```
{
  "semanticSchemaFileNames": ["CustomSemanticsSchema.json"]
}
```
that defines the schema file(s) that should be used for loading custom semantics definitions.

Some aspects of the implementation may still be polished and additional tests will be added.

### DIFF
--- a/specs/TilesetValidationSpec.ts
+++ b/specs/TilesetValidationSpec.ts
@@ -1,3 +1,4 @@
+import { ValidationOptions } from "../src/validation/ValidationOptions";
 import { Validators } from "../src/validation/Validators";
 
 describe("Tileset validation", function () {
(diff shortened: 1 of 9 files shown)


javagl:
The last commits consisted of minor cleanups and adding specs. 

The latter just add a [`customSemanticsSchema.json`](https://github.com/CesiumGS/3d-tiles-validator/blob/b35c662d2e64661121caba9655cd513bf08191d4/specs/data/tilesets/customSemanticsSchema.json) and two tilesets that refer to them - once correctly, and once with wrong types for the properties that receive these semantics. The specs then test that 

- a warning is generated for the unknown semantics when _not_ registering the custom semantics matcher
- no warning is generated when registering it
- errors are generated when the property types that receive the semantics don't match the definitions of the semantics themself

lilleyse:
Looks good, thanks @javagl

PR #330, opened 2025-01-31
Register empty validators for some MAXAR and VRICON extensions

bjornblissing:


javagl:
There is a part that [registers empty validators](https://github.com/CesiumGS/3d-tiles-validator/pull/330/files#diff-a78386a7231ad531334ce86445f92954eb17c385d2ebc5b34cbdf05df75cb283R488-R507) for `MAXAR_content_3tz` and `MAXAR_content_geojson`. This should not be necessary. The validators that are registered there are intended for the objects that are contained in the actual `extensions` dictionaries. But these extensions only allow certain content types. There are no actual  `extensions: { MAXAR_content_geojson: { ... } }` objects for which the validation should be skipped.

(Note: Maybe the intention here was to follow the pattern of [`3DTILES_content_gltf`](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_content_gltf). But the difference is: In this case, there **is** an `extensions: { 3DTILES_content_gltf: { ... } }` object that may contain additional information)

**But...** this is related to https://github.com/CesiumGS/3d-tiles-validator/issues/231 . Namely insofar that validating a tileset that declares `extensionsUsed: [ "MAXAR_content_geojson" ]` and actually _uses_ GeoJSON now (still) generates two warnings

- WARNING: Skipping 'geojson' validation
- WARNING: The extension 'MAXAR_content_geojson' was declared in 'extensionsUsed', but not found

I have some ideas for how this could be addressed, and will open a PR for this soon.

---

The other empty validators are for `VRICON_class` and `VRICON_grid`. And there, the intention is to suppress the
`WARNING: The extension 'VRICON_class' was used, but is not supported`.

However, I saw [`VRICON_class`](https://github.com/Maxar-Public/3d-tiles/tree/wff1.7.0/extensions/VRICON_class), but not `VRICON_grid`. Is it right to assume that `VRICON_grid` is some sort of "legacy" version of [`MAXAR_grid`](https://github.com/Maxar-Public/3d-tiles/tree/wff1.7.0/extensions/MAXAR_grid)? (If this is the case: `MAXAR_grid` already has an empty validator, and the approach that is implemented here would then indeed accomplish the goal of suppressing this message for `VRICON_grid` as well)

bjornblissing:
`VRICON_grid` is indeed the legacy version of `MAXAR_grid`.

Regarding `MAXAR_content_geojson`, it can either be used to indicate that a tile content is a `.geojson` file. Then there will be no object to validate, but it can also be used with a linked optional properties metadata schema for the GeoJSON content. Then there will be a extension object stored under the `metadataEntity.schema.json` in the tileset.
https://github.com/Maxar-Public/3d-tiles/blob/wff1.7.0/extensions/MAXAR_content_geojson/schema/metadataEntity.MAXAR_content_geojson.schema.json

Of course a better solution would be to actually validate this object as well.

javagl:
I see, I should have checked the spec of `MAXAR_content_geojson` more closely.

The handling of GeoJSON in general is a bit tricky: Right now, using `.geojson` triggers a warning
`"Skipping 'geojson' validation"`

It is not clear whether an "unknown/unhandled" content type should trigger a warning (or maybe just an info?). There is a [small comment](https://github.com/CesiumGS/3d-tiles-validator/blob/dcae5d305a639622a9bd48e8c622aa409b620790/src/validation/ContentDataValidators.ts#L89) in the respective code part about that, suggesting that it could make sense to make this configurable somehow.

(Note: Actually implementing a full-fledged GeoJSON validator is not in the scope of the 3D Tiles Validator. But of course, it could make sense to pull in a third-party validator for that. Quick searches suggest that there _are_ GeoJSON validation libraries, but I have not yet investigated which one could be the most suitable here).

I have created a follow-up PR for this one at https://github.com/CesiumGS/3d-tiles-validator/pull/331 

If you have any preferences about how to handle `.geojson` (i.e. whether it should generate a warning, info, or whether that should be configurable), just let me know - maybe this could be included in that PR as well.

PR #332, opened 2025-02-15
Update tools to 0.5.0

javagl:
(Builds on https://github.com/CesiumGS/3d-tiles-validator/pull/331 )

This updates the 3D Tiles Tools for the upcoming 0.5.0 release.

The main change for the (breaking) change to use the `async` API was pretty trivial - it just needed an `await` in a few places.

A minor change resulted from [this commit](https://github.com/CesiumGS/3d-tiles-tools/commit/5afdeab6a8435ed28149aee25bdbb1aa65b76cc2): Previously, the `TraversedTile` returned a 'subtree URI' for _both_ the _explicit_ tile that defined the implicit tiling, _and_ for the implicit root tile. Now, it only returns it for the implicit root tile. This should _usually_ not affect clients. But the validator is a different beast: It has to _explcitly_ access the `.subtree` file to perform an eager validation before even _starting_ the traversal.

A small change (that belongs to https://github.com/CesiumGS/3d-tiles-tools/commit/5afdeab6a8435ed28149aee25bdbb1aa65b76cc2 ) is: Previously, content types that are "known but not validated" (like `VCTR`, `GEOM`, and `GEOJSON`) caused a **WARNING** saying "Skipping _type_ validation". Now, this only is an **INFO**. The reasoning behind that is that the user can not do anything to "fix" this, so a **WARNING** does not seem warranted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ npx 3d-tiles-validator --optionsFile exampleOptions.json
 The options represent the properties of the `ValidationOptions` class. For example, using the following `exampleOptions.json` file, then the validator will only validate the tileset JSON structure, but _no_ tile content data:
 ```JSON
 {
-  "validateContentData": false,
+  "validateContentData": false
 }
 ```
 The following options will cause the validator to _include_ B3DM- and GLB files in the validation process, but ignore all other content types:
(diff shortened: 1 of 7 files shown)


javagl:
Note to myself: The `package.json` is not yet updated. I only tested this with a local (`file:...`) state, until the new 3D Tiles Tools version is released.

javagl:
This is now updated to use version 0.5.0 of the 3D Tiles Tools, which was just released.

Issue #335, opened 2025-05-28
False positive: EXTENSION_USED_BUT_NOT_FOUND


bjornblissing:
The following minimal example will emit the `EXTENSION_USED_BUT_NOT_FOUND` warning:

```json
{
  "asset": {
    "version": "1.1"
  },
  "geometricError": 0,
  "root": {
    "boundingVolume": {
      "region": [
        -1.31968,
        0.69887,
        -1.31960,
        0.69895,
        0.0,
        100.0
      ]
    },
    "geometricError": 0,
    "refine": "ADD",
    "content": {
      "uri": "unlit_tile.glb"
    }
  },
  "extensionsUsed": ["3DTILES_content_gltf"],
  "extensions": {
    "3DTILES_content_gltf": {
      "extensionsUsed": [
        "KHR_materials_unlit"
      ]
    }
  }
}
```

Log from validator:
```json
{
  "date": "2025-05-28T13:35:36.333Z",
  "numErrors": 0,
  "numWarnings": 1,
  "numInfos": 0,
  "issues": [
    {
      "type": "EXTENSION_USED_BUT_NOT_FOUND",
      "path": "/extensionsUsed",
      "message": "The extension '3DTILES_content_gltf' was declared in 'extensionsUsed', but not found",
      "severity": "WARNING"
    }
  ]
}
```

PR #336, opened 2025-06-02
Validation for `3DTILES_content_gltf`

javagl:
Supposed to be merged _after_ https://github.com/CesiumGS/3d-tiles/pull/810

Fixes https://github.com/CesiumGS/3d-tiles-validator/issues/335

The `3DTILES_content_gltf` extension is about to be un-deprecated via https://github.com/CesiumGS/3d-tiles/pull/810 . This means that it may now appear in 3D Tiles 1.1 data. There are some differences in terms of whether the extension is "used", "required", or "found" between 1.0 and 1.1. This is implemented in the [`TilesetValidator`](https://github.com/CesiumGS/3d-tiles-validator/blob/8484909a44fe35cabe9ce2881fd98d5092e3cef3/src/validation/TilesetValidator.ts#L340) as follows:

- For 1.0: 
  - When glTF content is encountered, then it must be declared in `extensionsUsed`
  - When the extension is in `extensionsUsed`, then it must also be in `extensionsRequired`
  - When the extension is in `extensionsUsed`, and no glTF content is encountered (and no extension object) then a warning will be emitted
- For 1.1:
  - Just encountering glTF content does **not** mean that it has to be in `extensionsUsed`!
  - Only when an actual `tileset.extensions["3DTILES_content_gltf"]` object is found, then it must be declared in in `extensionsUsed`
  - When it is declared as `extensionUsed`, but no extension object is found, then a warning will be emitted
  - (The extension is never _required_ in 1.1)

The actual `3DTILES_content_gltf { extensionsUsed ..., extensionsRequired }` objects are now undergoing the same consistency checks that have already been done for the _tileset_-level `extensionsUsed/Required` arrays. This functionality has been moved into a utility class and just called from the tileset validator and the new `ContentGltfValidator`. It checks whether the arrays have proper type, size, element types, unique elements, and whether the "required implies used" rule holds.

### DIFF
--- a/specs/data/extensions/contentGltf/contentGltfExtensionRequiredButNotUsed.json
+++ b/specs/data/extensions/contentGltf/contentGltfExtensionRequiredButNotUsed.json
@@ -1,0 +1,24 @@
+{
+  "extensionsUsed": ["3DTILES_content_gltf"],
+  "extensions": {
+    "3DTILES_content_gltf": {
+      "extensionsRequired": [
+        "KHR_materials_unlit"
+      ]
+    }
+  },  
+  "asset" : {
+    "version" : "1.1"
+  },
+  "geometricError" : 2.0,
+  "root" : {
+    "refine": "REPLACE",
+    "boundingVolume" : {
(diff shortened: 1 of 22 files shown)


lilleyse:
:+1:

Issue #318, opened 2024-11-08
Issue type for wrong enum value types


javagl:
A low-level detail, rather an internal note:

When there is an enum definition like
```
        "exampleEnum": {
          "type": "string",
          "enum": [
            "VALUE_A",
            "VALUE_B"
          ]
```
and the JSON input contains something like
`exampleEnum: 12345`
then this is reported as
`VALUE_NOT_IN_LIST:  The 'exampleEnum' property has the value 12345, but must be one of VALUE_A,VALUE_B`

This is correct and reasonable. 

One could make a case that this should be reported as a 
`TYPE_MISMATCH: The 'exampleEnum' property must have type 'string', but has type 'number'`

But I think that this is a special case, where the _type itself_ does not matter so much, and the details (i.e. the allowed values) add valuable information that justify using the more specific `VALUE_NOT_IN_LIST` issue.

PR #337, opened 2025-06-26
Add validation for MAXAR_grid extension

bjornblissing:
Implement full validation for `MAXAR_grid` extension based on `tileset.MAXAR_grid.schema.json` and `tile.MAXAR_grid.schema.json` specifications. The validator supports:

- Tileset-level validation for `quad`, `s2`, and `geod `grid types
- Tile-level validation for `boundingBox`, `level`, and `index `properties
- SRS validation with proper enum and pattern checking
- Automatic detection of tileset vs tile-level extensions
- Comprehensive test coverage with invalid test cases

Replaces the previous basic type-only validation with schema-compliant property validation for all `MAXAR_grid `extension properties.

### DIFF
--- a/specs/data/extensions/maxarGrid/invalidBoundingBoxSemantics.json
+++ b/specs/data/extensions/maxarGrid/invalidBoundingBoxSemantics.json
@@ -1,0 +1,28 @@
+{
+  "extras" : {
+    "info": {
+      "note": "Test file with invalid boundingBox semantics (xmin > xmax)"
+    }
+  },
+  "extensionsUsed": [
+    "MAXAR_grid"
+  ],
+  "extensions": {
+    "MAXAR_grid": {
+      "boundingBox": [100.0, 50.0, 50.0, 100.0, 0.0, 10.0],
+      "level": 5,
+      "index": [2, 3]
+    }    
+  },
+  "asset" : {
+    "version" : "1.1"
+  },
+  "geometricError" : 2.0,
+  "root" : {
+    "refine": "REPLACE",
+    "boundingVolume" : {
+      "box" : [ 0.5, 0.5, 0.5, 0.5, 0.0, 0.0, 0.0, 0.5, 0.0, 0.0, 0.0, 0.5 ]
+    },
+    "geometricError" : 1.0
+  }
+}
--- a/specs/data/extensions/maxarGrid/invalidCoordinateSystemType.json
+++ b/specs/data/extensions/maxarGrid/invalidCoordinateSystemType.json
@@ -1,0 +1,34 @@
+{
+  "extras" : {
+    "info": {
+      "note": "Test file with invalid coordinateSystem type (number instead of string)"
+    }
+  },
+  "extensionsUsed": [
+    "MAXAR_grid"
+  ],
+  "extensions": {
+    "MAXAR_grid": {
+      "type": "quad",
+      "center": [3097202.3706942615, 500000.00000000122],
+      "size": [2088960.0, 2088960.0],
+      "srs": {
+        "referenceSystem": "ITRF2008",
+        "epoch": "2005.0",
+        "coordinateSystem": 999,
+        "elevation": "ELLIPSOID"
+      }
+    }    
+  },
+  "asset" : {
+    "version" : "1.1"
+  },
+  "geometricError" : 2.0,
+  "root" : {
+    "refine": "REPLACE",
+    "boundingVolume" : {
+      "box" : [ 0.5, 0.5, 0.5, 0.5, 0.0, 0.0, 0.0, 0.5, 0.0, 0.0, 0.0, 0.5 ]
+    },
+    "geometricError" : 1.0
+  }
+}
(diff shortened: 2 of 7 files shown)


javagl:
Thanks for this contribution! 

<sub>(The task for adding validation for these extensions had been on my TODO list for quite a while, but only in the "When I have lots of time" section, which tends to be pushed down the list...)</sub>

Until now, I only had a slightly closer look at _this_ PR, but not yet at the counterpart at https://github.com/CesiumGS/3d-tiles-validator/pull/338 . The latter looks simpler, so I assume that whatever applies to _this_ one also applies there.

---

A high-level question in terms of goals and priorities: 

The validation seems to cover the relevant aspects. There are specs, and all tests are still passing. There is no strong reason to _not_ merge this as it is. 

So there are two options:

1. If your goal is to get this functionality in as quickly as possible, then I could merge these PRs (and create a new release) soon.
2. Otherwise, I could list a bunch of nitpicking comments

If you pick option **1.**, then you can ignore the following, and just add a comment saying "Merge it!!!" 🙂 

----

For option **2.**, here's the result of a first pass of nitpicking: 

- Adding the `extras.info.note` makes a lot of sense. It's sometimes hard to see what exactly is checked with the invalid spec files. I tried to encode this in the _file name_, but in some cases, the 'note' may make it clearer.

- The `validS2Grid.json` is not used/validated

- Until now, I tried to make the specs _exhaustive_. The goal was to cover _each and every aspect_ that is validated (or that can cause a file to be invalid) with a test case. To achieve that, it would be necessary to add a file that uses `MAXAR_grid: { type: "s2" }` with an additional `center` property, and one with a `size` property, and one with a `srs` property, to cover the [corresponding checks](https://github.com/CesiumGS/3d-tiles-validator/blob/5c152ceaa3cb63b3d5797ec0c08f29cd2efda4ea/src/validation/extensions/MaxarGridValidator.ts#L235). Similarly, it would be necessary to create files where the `MAXAR_grid.srs.coordinateSystem` is once `-999` (invalid type), and once `"UTM_INVALID"` (invalid value). (In addition to the valid `"UTM14N"` one). 
  Determining what _exactly_ can be valid or invalid can be difficult. For example, I'd have to closely read the specs to understand the cases that are covered with the `utmPattern` or so. 
  On top of that, there's the question of which "semantic" validations should be performed. One (random) example is that of `MAXAR_grid.boundingBox`, which is said to be `[xmin, ymin, xmax, ymax, zmin, zmax]` in [the specification](https://github.com/Maxar-Public/3d-tiles/tree/wff1.8.0/extensions/MAXAR_grid). And although the specification does not explicitly point that out as a requirement, I'd consider some validation that checks whether `xmin<=xmax`. (Should this hold? If so, it may be worth mentioning that in the spec...)

There are some places where the validation code could be simplified and/or made less verbose. Some examples:

- Checking whether several properties are not present, as in https://github.com/CesiumGS/3d-tiles-validator/blob/5c152ceaa3cb63b3d5797ec0c08f29cd2efda4ea/src/validation/extensions/MaxarGridValidator.ts#L235
could be done with `validateDisallowedProperties`, as in https://github.com/CesiumGS/3d-tiles-validator/blob/81403127b77f80428829dc2b31a02d3a646ec482/src/validation/extensions/NgaGpmValidator.ts#L502

- In some places, enums are first validated to be strings, and then their values are validated using the `validateEnum(...validValues)` pattern. For example, at https://github.com/CesiumGS/3d-tiles-validator/blob/5c152ceaa3cb63b3d5797ec0c08f29cd2efda4ea/src/validation/extensions/MaxarGridValidator.ts#L290 . The `validateEnum` does cover both. So strictly speaking, the type check is not necessary. But there's still the open question about what the desired validation error should be in these cases: https://github.com/CesiumGS/3d-tiles-validator/issues/318 . So both approaches (with and without explicit type check) can make sense.

- The part at https://github.com/CesiumGS/3d-tiles-validator/blob/5c152ceaa3cb63b3d5797ec0c08f29cd2efda4ea/src/validation/extensions/MaxarGridValidator.ts#L542 
could probably be done with `BasicValidator.validateIndexArray`. Similarly (below that), the `metersPerPixel` check could use `validateNumberRange`, and the `face` could use `validateIntegerRange`

(Admittedly, the "discoverability" of some of these convenience functions could be improved)

---

Again: I'm aware that this may appear to be nitpicking, but I see "nitpicking" as the _goal_ of the validator. And one can go _very_ far with things like this. If you wish (and this PR is not absolutely time-critical), then I could also create a PR _on top of this one_ that addresses some of these points.

javagl:
(Just to clarify: I did not thoroughly cross-check the implementation with the full specification - this might be out of scope for a review. And as an aside: Things like what I mentioned about the `coordinateSystem` would equally apply to all other properties, but I can see that this feels kind of tedious. I actually considered approaches for auto-generating test data for some of these cases. And there already are things like https://www.npmjs.com/package/json-schema-test-data-generator that could cover _some_ of that. The main problem would then be that it's not clear what _exactly_ is covered, and such a tool could not _explicitly_ cover the "semantic" validation issues anyhow). 

It looks like the last commit addresses the points that I mentioned, and I'll do another quick review/test pass on this and the other PR soon - as I said, I didn't see anything that would prevent it from being merged.

PR #341, opened 2025-07-03
Add MAXAR_image_ortho glTF extension validator

bjornblissing:
- Add MaxarImageOrthoValidator for validating MAXAR_image_ortho extensions in glTF images
- Validate transform property as 6-element number array (2x3 affine transformation matrix)
- Validate required srs property with comprehensive spatial reference system validation
- Support referenceSystem enum, epoch pattern, and complex coordinateSystem validation
- Handle UTM zone patterns, S2 cell patterns, and standard coordinate systems
- Validate optional properties with defaults (axis, unitHorizontal, unitVertical)
- Register validator in GltfExtensionValidators pipeline
- Add comprehensive test coverage for valid and invalid cases

Implements validation based on `image.MAXAR_image_ortho.schema.json` and `srs.schema.json` specifications for orthographically projected images with spatial reference metadata.


src/validation/extensions/maxar/MaxarValidatorCommon.ts
```diff
@@ -0,0 +1,246 @@
...
+    context: ValidationContext
+  ): boolean {
+    // Make sure that the given value is an object
+    if (!BasicValidator.validateObject(path, "srs", srs, context)) {
+      return false;
+    }
+
+    let result = true;
+
+    // Validate required properties
```
javagl, line 35: This is first checking whether all required properties are present. Then (below) it is checking each required property, but only if it is `defined(...)`.

I think that _this_ check here could be omitted. And instead, the checks below could all do their checks without the `if (defined(...))` wrapped around them.

(This should make no difference in behavior, but maybe it could be a bit simpler or clearer)

src/validation/extensions/maxar/MaxarValidatorCommon.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (defined(epoch)) {
+      const epochPath = path + "/epoch";
+      if (!BasicValidator.validateString(epochPath, "epoch", epoch, context)) {
+        result = false;
+      } else {
+        const epochPattern = /^[0-9]+(\.[0-9]+)?$/;
+        if (!epochPattern.test(epoch)) {
+          const message =
+            "The 'epoch' must match the pattern '^[0-9]+(\\.[0-9]+)?$'";
+          const issue = JsonValidationIssues.STRING_VALUE_INVALID(
```
javagl, line 84: Admittedly, there is some overlap between the meaning of `STRING_VALUE_INVALID` and `STRING_PATTERN_MISMATCH`. I think both are OK here.

src/validation/gltf/imageOrtho/MaxarImageOrthoValidator.ts
```diff
@@ -0,0 +1,162 @@
...
+        result = false;
+      }
+    }
+
+    // Validate the srs property (required)
+    const srs = maxarImageOrtho.srs;
+    const srsPath = path + "/srs";
+    if (!defined(srs)) {
+      const message = "The 'srs' property is required";
+      const issue = JsonValidationIssues.PROPERTY_MISSING(srsPath, message);
```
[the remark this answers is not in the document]
javagl: The following is obsolete when this is changed, but may be worth noting:

The `PROPERTY_MISSING` function is a "convenience" function that only expects the property _name_ as the second argument (and not the message). So right now, the message here becomes
`The 'The 'srs' property is required' property is required`

specs/gltfExtensions/MaxarImageOrthoValidationSpec.ts
```diff
@@ -0,0 +1,78 @@
+import { validateGltf } from "./validateGltf";
+
+describe("MAXAR_image_ortho extension validation", function () {
+  it("detects no issues in validMaxarImageOrtho", async function () {
+    const result = await validateGltf(
+      "./specs/data/gltfExtensions/maxarImageOrtho/validMaxarImageOrtho.gltf"
+    );
+    // Should only have info messages, no errors
```
javagl, line 8: Here (and in the tests below) there seems to be some "overhead" for diving into the result and analyzing its structure. 

I'll have to blame myself: Despite (usually) being pedantic, I did not properly document `validateGltf`.

The point is: `validateGltf` is **only** a utility function for validating the glTF _extensions_. It does not call the glTF Validator. So the results here should indeed _only_ contain the issues that are generated by the `MAXAR_image_ortho` validation. 

For illustration, logging each result of these specs with
`console.log(result.serialize());`
will print this:
```json
{
  "date": "2025-07-05T19:13:51.215Z",
  "numErrors": 0,
  "numWarnings": 0,
  "numInfos": 0
}
{
  "date": "2025-07-05T19:13:51.228Z",
  "numErrors": 1,
  "numWarnings": 0,
  "numInfos": 0,
  "issues": [
    {
      "type": "PROPERTY_MISSING",
      "path": "./specs/data/gltfExtensions/maxarImageOrtho/invalidMissingSrs.gltf/images/0/extensions/MAXAR_image_ortho/srs",
      "message": "The 'The 'srs' property is required' property is required",
      "severity": "ERROR"
    }
  ]
}
{
  "date": "2025-07-05T19:13:51.231Z",
  "numErrors": 1,
  "numWarnings": 0,
  "numInfos": 0,
  "issues": [
    {
      "type": "ARRAY_LENGTH_MISMATCH",
      "path": "./specs/data/gltfExtensions/maxarImageOrtho/invalidTransformLength.gltf/images/0/extensions/MAXAR_image_ortho/transform",
      "message": "Array 'transform' must have a length of 6, but the actual length is 4",
      "severity": "ERROR"
    }
  ]
}
{
  "date": "2025-07-05T19:13:51.233Z",
  "numErrors": 1,
  "numWarnings": 0,
  "numInfos": 0,
  "issues": [
    {
      "type": "STRING_VALUE_INVALID",
      "path": "./specs/data/gltfExtensions/maxarImageOrtho/invalidCoordinateSystem.gltf/images/0/extensions/MAXAR_image_ortho/srs/coordinateSystem",
      "message": "The 'coordinateSystem' must be 'GEOD', 'ECEF', match UTM pattern 'UTM(01-60)[NS]', or S2 pattern 'S2F[1-6]'",
      "severity": "ERROR"
    }
  ]
}
```
Where each contains exactly _one_ error that could be `expect`ed explicitly and dedicatedly.

src/validation/extensions/maxar/MaxarValidatorCommon.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // Check S2 pattern: S2F[1-6]
+    const s2Pattern = /^S2F[123456]$/;
+    if (s2Pattern.test(coordinateSystem)) {
+      return true;
+    }
+
+    // If none of the patterns match, it's invalid
+    const message =
```
javagl, line 173: Nitpick (?): I usually tried to be _very_ elaborate with the error messages. Specifically: I tried to _always_ include the actual value that was found. 

(A message like `"The value is wrong"` is far less helpful than `"The value must be X but was Y"`...)

javagl (commented):
Except for the linting warnings, this looks OK. One "convenient" way of fixing this could be to merge in https://github.com/CesiumGS/3d-tiles-validator/pull/344 and see whether CI fails then. 

Apart from that, the usual comment is that there may probably be additional configurations for failures that could be tested _explicitly_ (SRS with invalid epoch, invalid reference system, invalid elevation, etc...). But this is already in my notes.

src/validation/extensions/maxar/MaxarValidatorCommon.ts
bjornblissing: I have removed the `if (defined(...))` checks for each property below. But I still think that this test makes the validator a bit more readable IMHO.

javagl:
I think this should be ready for merge, except for the unused `JsonValidationIssues` import at https://github.com/CesiumGS/3d-tiles-validator/pull/341/files#diff-bca6da5b4ff9ff4ae7f86db9342ca32c7970e5f3b5634a62eb44d802dc79fc96R8

One suggestion would be to merge `main` into _this_ branch. This should include the https://github.com/CesiumGS/3d-tiles-validator/pull/344 , meaning that the build should fail afterwards (just to confirm that this works as expected). After removing the unused import, the build should pass, and then I'll merge it.

(You can also _only_ remove the unused import, the "test" for the linting fix PR can be done separately)

PR #342, opened 2025-07-03
Add MAXAR_temporal_light_traits glTF extension validator

bjornblissing:
- Add dedicated validator for MAXAR_temporal_light_traits extension that extends KHR_lights_punctual
- Validate flashing trait properties including waveform types, frequency, duty cycle, amplitude and phase
- Ensure extension is properly declared in extensionsUsed array
- Organize code with modular structure using MaxarTemporalLightTraitsValidator
- Include focused test suite with valid and invalid test cases
- Assume base KHR_lights_punctual validation is handled by glTF Validator

### DIFF
--- a/specs/gltfExtensions/KhrLightsPunctualValidationSpec.ts
+++ b/specs/gltfExtensions/KhrLightsPunctualValidationSpec.ts
@@ -1,0 +1,18 @@
+import { Validators } from "../../src/validation/Validators";
+
+describe("KHR_lights_punctual extension validation", function () {
+  it("detects no issues in validTilesetWithKhrLightsPunctual when no temporal traits are used", async function () {
+    const result = await Validators.validateTilesetFile(
+      "specs/data/gltfExtensions/khrLightsPunctual/validTilesetWithKhrLightsPunctual.json"
+    );
+    // Should only have info messages, no errors since we only validate temporal traits
+    let hasErrors = false;
+    for (let i = 0; i < result.length; i++) {
+      if (result.get(i).severity === "ERROR") {
+        hasErrors = true;
+        break;
+      }
+    }
+    expect(hasErrors).toBe(false);
+  });
+});
--- a/specs/gltfExtensions/KhrLightsPunctualValidationSpec.ts
+++ b/specs/gltfExtensions/KhrLightsPunctualValidationSpec.ts
@@ -6,13 +6,17 @@ describe("KHR_lights_punctual extension validation", function () {
       "specs/data/gltfExtensions/khrLightsPunctual/validTilesetWithKhrLightsPunctual.json"
     );
     // Should only have info messages, no errors since we only validate temporal traits
-    let hasErrors = false;
-    for (let i = 0; i < result.length; i++) {
-      if (result.get(i).severity === "ERROR") {
-        hasErrors = true;
-        break;
-      }
-    }
-    expect(hasErrors).toBe(false);
+    expect(result.length).toEqual(0);
+  });
+
+  it("detects issues in validTilesetWithInvalidKhrLightsPunctual", async function () {
+    const result = await Validators.validateTilesetFile(
+      "specs/data/gltfExtensions/khrLightsPunctual/validTilesetWithInvalidKhrLightsPunctual.json"
+    );
+    // Expect one error from the glTF Validator that complains about
+    // the MAXAR_temporal_light_traits extension being used within
+    // one light, but not being declared in extensionsUsed.
+    expect(result.length).toEqual(1);
+    expect(result.get(0).type).toEqual("CONTENT_VALIDATION_ERROR");
   });
 });
--- a/specs/gltfExtensions/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidationSpec.ts
+++ b/specs/gltfExtensions/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidationSpec.ts
@@ -1,0 +1,144 @@
+import { validateGltf } from "../validateGltf";
+
+describe("MAXAR_temporal_light_traits extension validation", function () {
+  it("detects no issues in validTemporalLightTraits", async function () {
+    const result = await validateGltf(
+      "./specs/data/gltfExtensions/khrLightsPunctual/maxarTemporalLightTraits/validTemporalLightTraits.gltf"
+    );
+    // Should only have info messages, no errors
+    let hasErrors = false;
+    for (let i = 0; i < result.length; i++) {
+      if (result.get(i).severity === "ERROR") {
+        hasErrors = true;
+        break;
+      }
+    }
+    expect(hasErrors).toBe(false);
+  });
+
+  it("detects missing waveform property in temporal light traits", async function () {
+    const result = await validateGltf(
+      "./specs/data/gltfExtensions/khrLightsPunctual/maxarTemporalLightTraits/invalidMissingWaveform.gltf"
+    );
+    expect(result.length).toBeGreaterThan(0);
+
+    // Find the error about missing waveform property
+    let foundMissingWaveform = false;
+    for (let i = 0; i < result.length; i++) {
+      const issue = result.get(i);
+      if (
+        issue.type === "INVALID_GLTF_STRUCTURE" &&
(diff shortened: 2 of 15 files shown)


specs/gltfExtensions/KhrLightsPunctualValidationSpec.ts
```diff
@@ -0,0 +1,18 @@
+import { Validators } from "../../src/validation/Validators";
+
+describe("KHR_lights_punctual extension validation", function () {
+  it("detects no issues in validTilesetWithKhrLightsPunctual when no temporal traits are used", async function () {
+    const result = await Validators.validateTilesetFile(
```
javagl, line 5: While it might not be strictly necessary to validate a _tileset_ file that contains the valid glTF, I think that this makes sense here. I makes sure that the validator does properly handle a tileset with `KHR_lights_punctual` content (and does not generate issues/warnings for that). Particularly because the `MaxarTemporalLightTraitsValidator` is "nested" in the `KhrLightsPunctualValidator`. So this establishes the baseline on which the `MaxarTemporalLightTraitsValidationSpec` builds.

But also see the note about the `extensionsUsed` validation below, and in the final/summarizing review comment!

src/validation/gltf/lightsPunctual/KhrLightsPunctualValidator.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    let result = true;
+
+    // Validate MAXAR_temporal_light_traits extension if present
+    const extensions = light.extensions;
+    if (defined(extensions)) {
+      const temporalTraits = extensions["MAXAR_temporal_light_traits"];
+      if (defined(temporalTraits)) {
+        if (!hasTemporalTraits) {
+          const message =
```
javagl, line 110: This is checked by the glTF validator. See the notes in the final/summarizing review comment.

src/validation/gltf/lightsPunctual/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidator.ts
```diff
@@ -0,0 +1,197 @@
...
+      return false;
+    }
+
+    let result = true;
+
+    // Validate the waveform property (required)
+    const waveform = flashing.waveform;
+    if (!defined(waveform)) {
+      const message = "The 'waveform' property is required";
+      const issue = GltfExtensionValidationIssues.INVALID_GLTF_STRUCTURE(
```
javagl, line 83: I've seen that all/most of these are reporting a `INVALID_GLTF_STRUCTURE`. I think that most of them could/should use the same issue structures and mechanisms as the baseline 3D Tiles validator. For example, the `defined` check here and the manual `validWaveforms` check could be handled with the `validateEnum` function.

src/validation/gltf/lightsPunctual/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidator.ts
```diff
@@ -0,0 +1,197 @@
...
+        );
+        context.addIssue(issue);
+        result = false;
+      }
+    }
+
+    // Validate the frequency property (required)
+    const frequency = flashing.frequency;
+    if (!defined(frequency)) {
+      const message = "The 'frequency' property is required";
```
javagl, line 107: Could be a simple `BasicValidator.validateNumberRange`

specs/gltfExtensions/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidationSpec.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(hasErrors).toBe(false);
+  });
+
+  it("detects missing waveform property in temporal light traits", async function () {
+    const result = await validateGltf(
+      "./specs/data/gltfExtensions/khrLightsPunctual/maxarTemporalLightTraits/invalidMissingWaveform.gltf"
+    );
+    expect(result.length).toBeGreaterThan(0);
+
+    // Find the error about missing waveform property
```
javagl, line 25: Such inconvenient checks should not be necessary. See final/summarizing review comment

javagl (commented):
Some higher-level comments:

Some aspects of the glTF extensions validation within the 3D Tiles Validator may have to be clarified (and better documented). And I have taken an internal TODO note for that.

One specific example:

I've seen that you manually checked for the `extensionsUsed` when validating the `MAXAR_temporal_light_traits` **within** the `KHR_lights_punctual`. And I can see where this came from: When you are using the `validateGltf` utility function from the specs, then this will **only** perform the validation that is done by the 3D Tiles Validator. It will **not** call the glTF Validator. This means that in order to receive error messages for the `invalidUndeclaredExtension.gltf`, you seemingly _have_ to do this manual check for the `extensionsUsed`. But this is not necessary when the _full_ glTF validation is performed. 

As an example, I created https://github.com/CesiumGS/3d-tiles-validator/commit/8d9eb02945df774122e809bc7117919161fb9e0f : It embeds the `invalidUndeclaredExtension.gltf` into a tileset, and then performs a default 3D Tiles Validator run on that tileset. This **does** include the glTF validator call, meaning that it will already bail out with an error, because the glTF Validator detects the missing `extensionsUsed`.

---

One reason for why the `validateGltf` function does **not** call the glTF validator is that it should be possible to "focus" on the part that the 3D Tiles Validator is responsible for, and not having to deal with the "noise" (that might be generated by the glTF validator) when defining the specs.

This is related to the comment about the elaborate and inconvenient checks that may have _seemed_ to be necessary for the specs here. For example, the check for the `invalidMissingWaveform`. Given that `validateGltf` **only** performs the validation that is implemented as part of the 3D Tiles validator, the output there does _not_ have to be expected to be polluted by glTF Validator noise. It can be expected to **only** contain that single issue, which is the ERROR about the invalid `waveform`.

---

It's not perfectly clear how to improve that. As mentioned in the previous review: The `validateGltf` was only intended as an internal utility function, and not sufficiently documented. 

One could consider to differentiate that utility function into (roughly) something like

- `validateGltfFull` - To actually run the glTF validator and _then_ the 3D Tiles Validator (glTF) validation steps
- `validateGltf3DTilesOnly` - To **only** perform the 3D Tiles Validator glTF validation steps (and to avoid noise that may come from the glTF Validator)

Do you think that something like this could make sense?

---

Another general remark: 

I think that letting all glTF errors boil down to `INVALID_GLTF_STRUCTURE` is losing some valuable information. It may make the messages less clear and consistent, and make it harder to (for example) filter out or search for certain _specific_ issues _within_ the glTF issues.  As mentioned in the inlined comment, I think that the code could be simplified (and the output could be more consistent) if these used the standard, built-in functions, like `BasicValidator.validateNumberRange` and such.

src/validation/gltf/lightsPunctual/MaxarTemporalLightTraits/MaxarTemporalLightTraitsValidator.ts
```diff
@@ -0,0 +1,190 @@
...
+    // Validate the duty property (optional)
+    const duty = flashing.duty;
+    if (defined(duty)) {
+      if (
+        !BasicValidator.validateNumberRange(
+          path + "/duty",
+          "duty",
+          duty,
+          0.0,
+          true,
```
javagl, line 121: A nitpick that rather refers to the _specification_: 

The descriptions at https://github.com/Maxar-Public/glTF/tree/wff1.8.0/extensions/2.0/Vendor/MAXAR_temporal_light_traits#flashing say

> ... as normalized value `(0.0, 1.0)`, only ...

The brackets _usually_ indicate whether the start and end of a range are inclusive. For example,

`[0,1]` both inclusive
`(0,1)` both exclusive
`[0, 1)` including 0, excluding 1
`(0, 1]` excluding 0, including 1

(This is also pretty consistently applied in other parts of glTF and 3D Tiles specifications)

In terms of the validation: According to the schema, the range is indeed _inclusive_, meaning that the `true` here is correct (and the range should probably be denoted as `[0,1]` in the spec text)

javagl (commented):
Apart from the unused import, this should be ready to merge

Issue #233, opened 2022-10-17
Validate that geometry is fully contained in bounding volumes


javagl:
One aspect of the validation of content data should be to check that geometry is fully contained in the relevant bounding volumes. This refers to `content.boundingVolume` and each `tile.boundingVolume` on the path from the tile that contains the geometry, up to the root.

javagl:
A small update here: When this issue was opened, it would have been a lot of effort to actually implement this, because there was no functionality for computing the bounding volume of the tile content in the first place. In the meantime, there are some functionalities in the `3d-tiles-tools`, surrounding the https://github.com/CesiumGS/3d-tiles-tools/blob/ac192defb95fde7c93a06897bf68228bd3ad87ce/src/tools/tilesetProcessing/ContentBoundingVolumes.ts#L25 class, which may allow implementing this check with reasonable effort. 

It would likely be computationally expensive, so should probably be disabled by default via a validation config flag. And it could be difficult to make this check "perfect" - for example, some _really_ clever and expensive implementation could compute a bounding volume for an I3DM that is _tighter_ than the one that is computed with the 3D Tiles Tools. But it could at least be implemented as a preview feature, maybe even with some "threshold" so that small deviations would only be a WARNING. In any case, the building blocks are now there.

javagl:
An aside: A seemingly straightforward optimization of that last approach would be to check whether the geometry is contained in the leaf bounding volume, and then first only check whether the leaf bounding volume is fully contained in its parent (up to the root). If this is the case, then it will imply that the geometry is also fully contained in each of the parent bounding volumes. But 1. when this is not the case, the low-level, really geometry-based check has to be done, and 2. one has to think about how to carry that information up the hierarchy.

javagl:
Geometry that does not match the bounding volumes also seems to be the reason for the issue behind https://community.cesium.com/t/does-uploading-a-3d-tile-to-ion-change-the-mesh-at-all/39058

javagl:
Only a note: There are currently containment checks for geometry in bounding regions as part of https://github.com/CesiumGS/3d-tiles-validator/pull/347 . These generate a `BOUNDING_VOLUMES_INCONSISTENT`. (There is no better issue type for now). Once the geometry validation is addressed more broadly, there will likely be a special issue type, something like `GEOMETRY_NOT_IN_BOUNDING_VOLUME` or so. Once this is addressed, some of the issue types may be updated.

javagl:
[...]

PR #347, opened 2025-08-20
Enhanced MAXAR_extent Extension Validation

bjornblissing:
Enhances the `MAXAR_extent` extension validator with URI resolution check, GeoJSON validation, and spatial containment checking. 

- Validates that the `uri` property points to a resolvable resource
- Integrates with `GeojsonValidator` to ensure referenced content is valid GeoJSON
- Enforces that extent GeoJSON contains only `Polygon` or `MultiPolygon` geometries
- Ensures polygon rings have at least 3 unique coordinates
- Handles ring closure and duplicate coordinate detection
- Implements line segment intersection algorithm to detect and forbid self-intersecting polygons
- Validates that all GeoJSON coordinates fall within the root tile's bounding volume
  - Region Support: Full support for region-based bounding volumes with proper coordinate system handling
  - _Future Extensibility:_ Architecture supports box and sphere bounding volumes


javagl:
I only quickly skimmed over the changes for now. There will probably be a few more detailed review comments soon. Only bullet points (no _specific_ actions/change requests for now): 

- There are some functions and parameters that are not documented
- There was a comment "Skip invalid geometries (already handled by GeoJSON validator)". In general, when the validator detects that 'something' is invalid, it does to validate 'as much as possible', but skip what cannot be validated. This does raise the question: When exactly does it have to bail out with a `return false`? There are some places where it differentiates between a "JSON-level validity" and a "higher-level validity". In this case, I'll have a closer look at whether it could make sense to differentiate that similarly, to check whether the JSON of geometry is valid, and (only if it is) do the additional, geometric validity checks
- There seem to be several places where some of these tricky "epsilon"-questions come into play. The insersection checks are using 1e-10. The computation of unique positions converts values to strings, and that would treat 0.4000000000000000021 and 0.4000000000000000022 as different values (and I don't know whether this causes trouble at other places, where real epsilons are used...)

There are the `TODO`s for the sphere/box validation. It's not clear how this could be resolved. (Maybe just a console message like "That's a box/sphere, not checking that yet" or so - TBD). 

But a bit more generally, about containment checks: There are still pending issues https://github.com/CesiumGS/3d-tiles-validator/issues/25 and https://github.com/CesiumGS/3d-tiles-validator/issues/233 that are in a similar ballpark. I recently started to work on some of that (locally, not a public branch yet). This happened in the `3d-tiles-tools`. But the idea there was to generalize some of the functionality that already _is_ contained in the 3D Tiles tools: They already offer the functionality to _compute_ bounding volumes from tile content. And it's only a small step to generalize this to a function that just _iterates_ over all points of a geometry, and passes them to a "consumer". This consumer could then use them to still compute a bounding volume (in the tools). But this consumer could also be the validator, that then _checks_ if the given positions _are_ in a _given_ bounding volume. 

The different nature of `box/sphere` vs. `region` (and the conversion between cartographic and cartesian) do raise a few questions about the exact architecture for that. 

I'll try to do a thorough review ASAP.

src/validation/extensions/MaxarExtentValidator.ts
```diff
@@ -98,9 +104,637 @@ export class MaxarExtentValidator implements Validator<any> {
...
+      checkGeometry(geojsonObject);
+    }
+
+    // Report any invalid geometry types found
+    if (invalidGeometries.length > 0) {
+      const uniqueTypes = [...new Set(invalidGeometries)];
+      const message = `MAXAR_extent GeoJSON must contain only Polygon or MultiPolygon geometries, but found: ${uniqueTypes.join(
+        ", "
+      )}`;
+      const issue = SemanticValidationIssues.BOUNDING_VOLUMES_INCONSISTENT(
```
javagl, line 331: This (as well as the checks below in `validatePolygonExtent`) report a `BOUNDING_VOLUMES_INCONSISTENT`. This issue currently has a _very_ narrow and specific meaning, and does not really fit here. 

Given that the validation is (currently) very specific for GeoJSON, and the type of validation that is performed here (sort of a "geometric" validation) is new in the context of the validator, I'd suggest to introduce a new issue type for that.

There question about "how specific" the issue types should be is something that I've been pondering with. There is no silver bullet. It may be subjective. The right answer is somewhere between `SOMETHING_WRONG` and `PROPERTY_X_IS_NOT_FORTY_TWO`. 

A first shot:

- There could be a `GeometricValidationIssues` class
- There could be `INVALID_GEOMETRY_TYPE` for the issue _here_
- There could be `INVALID_GEOMETRY_SIZE` (or so) for the 'ring with < coords) (below)
- There could be `INVALID_GEOMETRY_STRUCTURE` (or so) for the 'self-intersections' (below)

I'm open for other suggestions or granularities, but think that ~"something like that" could make sense...

src/validation/extensions/MaxarExtentValidator.ts
```diff
@@ -98,9 +104,637 @@ export class MaxarExtentValidator implements Validator<any> {
...
+        ring[0],
+        ring[ring.length - 1],
+        epsilon
+      )
+        ? ring.slice(0, -1)
+        : ring;
+
+    for (const coord of coordsToCheck) {
+      // Check if this coordinate is already in the unique list
+      const isDuplicate = unique.some((existingCoord) =>
```
javagl, line 485: Only a remark, no change request: This has O(n²) complexity. This _may_ become an issue for large geometries. But given the epsilon-handling, there is no efficient way for solving that (except for hyper-complex spatial data structures).

src/validation/extensions/MaxarExtentValidator.ts
```diff
@@ -98,9 +104,637 @@ export class MaxarExtentValidator implements Validator<any> {
...
+   */
+  static doLineSegmentsIntersect(
+    p1: number[],
+    q1: number[],
+    p2: number[],
+    q2: number[]
+  ): boolean {
+    const orientation = (p: number[], q: number[], r: number[]): number => {
+      const val = (q[1] - p[1]) * (r[0] - q[0]) - (q[0] - p[0]) * (r[1] - q[1]);
+      if (Math.abs(val) < 1e-10) return 0; // Collinear
```
javagl, line 574: Admittedly, I didn't do the math behind that that intersection function. The math always boily down to the same thing, but there are many different ways of actually _implementing_ it, and I _assume_ that this is a "standard" way of doing this.

Just a detail, regarding that `1e-10`: This is the same value as the `epsilon` that is defined in `getUniqueCoordinates`. I think that there are roughly (!) two options for stuff like that:

- One could pass in an `epsilon` parameter everywhere (i.e. passing it in to `getUniqueCoordinates` and handing it from `isRingSelfIntersecting` to `doLineSegmentsIntersect`). This can feel clumsy...
- One could define it _once_ as a `private static readonly EPSILON = 1e-10;` for the whole class

Given that the functions in this class are nearly all `private` ...

(or ... they _should_ be `private`... that's one thing that may be fixed as well...)

... I think that defining a static `EPSILON` that is used throughout that class is reasonable here. **Only** when functions are supposed to be called from outside, an actual epsilon _parameter_ is really required (and this is just not the case here right now)

javagl (commented):
Overall, this looks good. There are a few minor inlined comments.

As mentioned in previous PRs: For some aspects of the GeoJSON validation, I'd have to read and digest the full GeoJSON spec first, to better understand what exactly is checked (or what _might_ be missing). But from reading the validation code itself, with the different 'types' and the 'recursive' handling of `GeometryCollection` and such, it looks like this all makes sense.

---

Notes to myself (**no** change requests!)

- Most of the `validateObject` functions do not carry a JSDoc comment. (I noticed that in the `GeojsonValidator` - although I _usually_ try to be pedantic, and apply the same scrutiny to my own code as to the code that I'm reviewing, that one's on me...)
- There are several places in the code (not only in this PR!) where JSON is parsed from a `Buffer`. I'll consider to wrap that into some utility function, to handle that more conveniently and have ~"a central place for generating the `JSON_PARSE_ERROR`"...
- The error for a point not being in the bounding volume is `BOUNDING_VOLUMES_INCONSISTENT`. There currently is no better issue type for that. I took a note at https://github.com/CesiumGS/3d-tiles-validator/issues/233#issuecomment-3218112153 that this _may_ become some more specialized issue type in the future. (There's an inlined comment about other cases where `BOUNDING_VOLUMES_INCONSISTENT` was used, but that one is independent)

Issue #317, opened 2024-10-19
Consider filtering out obsolete infos from glTF validator


javagl:
The 3D Tiles Validator currently supports the validation of several glTF extensions. Under the hood, this is implemented by

- using the glTF Validator to perform the baseline validation
- dive into the glTF to perform the extension-specific validation

Calling the glTF-Validator with a glTF that contains an unknown extension will still cause an 'INFO' message saying

> "Cannot validate an extension as it is not supported by the validator: '...'."

These messages could be fitered out by default for the extensions that are validated by the 3D Tiles Validator. On this level, this could be pretty simple.

A slightly more challenging aspect of this could be to also filter out the messages that refer to potentially unused elements. When a certain texture or accessor or attribute is _only_ used in an extension, then the glTF Validator will report
```
                  "type": "CONTENT_VALIDATION_INFO",
                  "path": "/textures/3",
                  "message": "This object may be unused.",
                  "severity": "INFO"
```

The goal would then be to filter out this message **if and only if** the object is used in one of the known extensions. This is _technically_ possible, by looking at the `path` and matching this against the elements in the glTF that have been found to be _used in the extension_. But transporting that information back from the validation process and applying it as a filter to the output that is generated by the glTF-Validator may require some thought.

javagl:
The part of filtering out the

> "Cannot validate an extension as it is not supported by the validator: '...'."

messages should be simple. And it could be important, because for end-users, it's not obvious that this message is coming from the glTF-Validator, and not the 3D Tiles Validator. (Specifically: They might assume that the 3D Tiles Validator is emitting this message, **even though** it does support and validate a specific extension).

Filtering out the infos about potentially unused elements is more challenging, but could the addressed separately, and should not prevent the simple and useful part from being implemented.

javagl:
Both messages are generated by the glTF-Validatorm, and if validation for `KHR_texture_basisu` was added, then _both_ of them could be filtered out in the 3D Tiles Validator.

But I have to start asking at a much higher level: 

What is the intended scope of this validation? 

The 3D Tiles Validator is meant for validating 3D Tiles. It includes a validation for glTF, because glTF is the main payload for 3D Tiles (even when it is wrapped into B3DM or so), and glTF Validator offers a nice library functionality that can easily be integrated here.

[...]

PR #348, opened 2025-08-30
Allow filtering out glTF Validator issues

javagl:
Addresses https://github.com/CesiumGS/3d-tiles-validator/issues/317

This is a **DRAFT**, to illustrate some possible approaches.

---

The first part of the current state (commit https://github.com/CesiumGS/3d-tiles-validator/commit/e96ed3289c104028c40fbe1d9e5c3848a6483c30) is a minor generalization of the `GltfExtensionValidators` class. 

Until now, this class only offered a function called `validateGltfExtensions`. This did call the glTF extension validators that have been implemented within the 3D Tiles Tools. It simply did call them, one by one, by name. 

The generalization here is: It now registers these validators in an `extensionName -> GltfExtensionValidator` lookup. This allows the implementation to simply iterate over this lookup and call them generically. 
More importantly: This also allows the class to offer a function `isRegistered(extensionName: string) : boolean` that allows checking whether a certain extension is implemented as part of the 3D Tiles Validator (this will be used in the filtering - see below)

---

The second part (commit https://github.com/CesiumGS/3d-tiles-validator/commit/f11875f640867f17103f56a1740f74f9baddd66a ) is the **very** drafty part for filtering out certain issues that are generated by the glTF Validator.

Some context: 

The 3D Tiles Validator calls the glTF Validator for each relevant tile content. The glTF Validator generates a validation report. This validation report is translated into the `ValidationIssue`-based format of the 3D Tiles Validator. So the glTF Validator output ends up as the list of "causes" for a `CONTENT_VALIDATION_ERROR/WARNING/INFO`. 

The 3D Tiles Validator already _does_ offer functionality for filtering validation results. So one could, in theory, define some `ValidationIssueFilter` to filter out the unwanted glTF Validator issues from the _final_ `ValidationResult`. The reasons why I did not choose this option:

- Depending on how many issues are generated by the glTF-Validator, this may cause the final `ValidationResult` to become **huge**. I think that it can make sense to filter out unwanted warnings at an earlier place - namely, where they are _created_. 
- Related to that: Some warnings that may have to be filtered out are very "context specific". Trying to filter them out, post-hoc, from the full `ValidationResult` could 1. involve very complex `ValidationIssueFilter` conditions, and 2. more importantly: make it necessary to derive that context information by ~"searching other messages" in the full `ValidationResult`

So what is done here now:

The issues that are generated by the glTF Validator are translated into `ValidationIssue` objects. These will become the "causes" of the corresponding `CONTENT_VALIDATION_ERROR/WARNING/INFO` of the 3D Tiles Validator.

There is now a **draft** function [`filterCauses`](https://github.com/CesiumGS/3d-tiles-validator/blob/f11875f640867f17103f56a1740f74f9baddd66a/src/tileFormats/GltfValidator.ts#L250) that receives these validation issues, for **one** glTF asset, and can implement different filter criteria.

One of the filter criteria is that the issues with the message
`"Cannot validate an extension as it is not supported by the validator: '"`
are filtered out for all extensions that are "registered" according to the `GltfExtensionValidators.isRegistered(name)` function.

A **dummy** implementation of [certain filter criteria for `KHR_texture_basisu`](https://github.com/CesiumGS/3d-tiles-validator/blob/f11875f640867f17103f56a1740f74f9baddd66a/src/tileFormats/GltfValidator.ts#L281) is shown in the current state: It filters out issues 
- about the unsupported extension itself
- about the invalid media/MIME type `image/ktx2`
- about unrecognized image formats

Note that these are just dummy implementations. For example, the `"Image format not recognized."` message could also be caused for WEBP images! A proper filter would have to inspect the glTF, and see whether this message is really caused by a KTX texture.


javagl:
An example for how that fitering infrastructure could be applied is now implemented (as a draft) for `EXT_structural_metadata` and `EXT_mesh_features`.

There is a function [`processCauses`](https://github.com/CesiumGS/3d-tiles-validator/blob/7f042e7637d2beb9b4d22cc799b2075deec09e3a/src/validation/gltf/structuralMetadata/ExtStructuralMetadataIssues.ts#L24) for this extension. This function is simply used for the interface implementation when the corresponding validator is [registered](https://github.com/CesiumGS/3d-tiles-validator/blob/7f042e7637d2beb9b4d22cc799b2075deec09e3a/src/validation/gltf/GltfExtensionValidators.ts#L154).

This function will be called with the list of causes/issues that are generated by the glTF validator, and return the filtered result. It will 

- remove the issue about the `EXT_structural_metadata` extension not being supported
- remove all issues that say "This object may be unused" for a buffer view or texture that actually **is** used by the extension

The latter is solved pragmatically: There are functions that compute the indices of buffer views or textures that **are** used, by drilling through the raw glTF JSON (defaulting to `{}` or `[]` everywhere), and returning the list of used indices. A utility function
`GltfExtensionIssues.isObsoleteIssueAboutUnusedObject(issue, "bufferViews", usedBufferViewIndices);`
checks whether the issue is such a message, allowing to filter it out accordingly.

PR #351, opened 2025-09-02
MAXAR_extent: Fix intersection test

bjornblissing:
**Improvement:**
- Detect and report intersecting edge indices and intersection point
- Indicate when the intersection occurs at an existing vertex

**New Tests**
* near-miss (no intersection, should pass)
* true crossing (should intersect. should fail)
* collinear overlap (invalid geometry, should fail)
* Overlapping rings without self-intersections (should pass)


bjornblissing:
> I think that the current tests are so specific that they cannot be derived from this statement. They do involve some "knowledge" about what should be "valid" and "invalid", and this knowledge doesn't seem to be formalized anywhere - or is it?

There are work ongoing to improve the language of the `MAXAR_extent` extension. This work is currently under internal review. There are no changes planned that affects already existing datasets. But we are trying to iron out unclear or ambiguous language.

javagl:
Sounds good. One point here is, a bit vaguely: When there is a statement that "extents may not self-intersect", then that's something that might "only" be relevant for a renderer, meaning that the renderer does not have to care about some "winding order" that causes certain areas to be filled or not filled, and even if there's an epsilon-sized self-intersection, this may not be noticable (unless the user zooms in, up to another epsilon distance, to _this exact spot_). 

For the validator, this may be important. Consider the left edge being at x=0.0, and the circled vertex here:

Cesium SelfIntersect

Does it "self-intersect" when the circled vertex is at 

- x = -1e-10 ?
- x = 0.0 ?
- x = +1e-10 ? 

I didn't use pen+paper to draw the GeoJSON test cases yet (but ... eventually, I _might_ actually do this 😆 ). 

As mentioned in an earlier PR: The epsilons can be tricky. When computing the (signed) distance of that vertex to the edge, with the **theoretical** cases of `d<0`, `d=0`, `d>0`, then the case of `d=0` will never happen _in practice_ (except for artificial cases like that exactly vertical line). It's hard to say how (strictly) this should be carved out in the actual specification text. (It may involve some "domain knowledge" about which computations are done _exactly_ with this geometry, and how lenient that computation is for such epsilon-cases...)

bjornblissing:
Proposed updated language:
>- **Self-Intersection**: Individual polygons and multipolygons must not self-intersect. A polygon self-intersects when its boundary (ring) crosses or touches itself at any point other than at vertices that are shared between consecutive edges. This includes cases where edges cross each other, creating figure-eight patterns, or where the ring touches itself creating invalid topological structures

javagl:
I think that the ... "crosses **or touches**" makes clear that the `x=0` case should count as a self-intersection. 

(As usual, the epsilon handling could mean that _the implementation_ counts a case like `x=1e-123` as "touching" as well (because that _is_ `x=0` "for all intents and purposes"), but this cannot be covered by the spec...)

PR #352, opened 2025-09-04
MaxarValidatorCommon: correct S2 face range

bjornblissing:
Correct S2 face range to 0–5 in MaxarValidatorCommon


javagl:
Looks simple and reasonable, but I'd like to confirm: https://github.com/Maxar-Public/glTF/blob/6262a06b4b865c653bfbd274deb83dcb317e21ea/extensions/2.0/Vendor/MAXAR_image_ortho/schema/srs.schema.json#L30 looks like a different pattern. (It's late here, maybe I overlooked something obvious - in doubt, I'll take a closer look soon)

